PR #5, opened 2018-06-27
Initial PR for Azure SB over amqp

amarzavery:
This PR contains
- common code in lib/amqp-common. Stuff here will be reused between EH andSB.
- Promisified rhea layer in lib/rhea-promise. Several changes have been made in lib/rhea-promise from what was done in event hubs.
- SB code in lib


amarzavery:
I would like to get the design sorted for things like
- what should happen in peekLock and receiveAndDelete mode
- whether we should have a prefetchCount option? Currently I do not have that for a reason. I believe we are better off getting  messages from the wire, rather than prefetching them in an internal queue and doing more processing with it to make sure that the lock on them does not expire.
- The QueueClient has two methods for receiving messages
  - receive() method takes in the message and the error handler and some optional parameters. There will be one dedicated receive link for receiving messages via handlers per queue client.
  - receiveBatch() takes in the max message count and max wait time. It returns an array of messages depending on whether maximum messages were received or the wait time out happened. There will be one dedicated receive link for receiving messages in a batch per queue client.
- The Queue client has two methods for sending messages
  - send() and sendBatch(). There will be one sender link per queue client.

.vscode/settings.json
```diff
@@ -0,0 +1,26 @@
+{
```
princjef: nit: do all of these settings need to be in the repo settings? not all users do things the same way and you can achieve this for all of your projects by putting the same information in your user settings

package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@types/chai-as-promised": "^7.1.0",
+    "@types/mocha": "^5.0.0",
+    "chai": "^4.1.2",
+    "chai-as-promised": "^7.1.1",
+    "mocha": "^5.0.5",
+    "ts-node": "^5.0.1",
+    "tslint": "^5.9.1",
+    "typescript": "^2.9.1",
+    "dotenv": "^5.0.1",
+    "@types/dotenv": "^4.0.3",
```
princjef: nit: let npm modify the dependency list for you through `npm i <package>` (only reason i can tell is because it automatically sorts all dependencies each time)

package.json
```diff
@@ -0,0 +1,53 @@
...
+    "nyc": "^11.8.0"
+  },
+  "scripts": {
+    "tslint": "tslint -p . -c tslint.json --exclude examples/**/*.ts --exclude test/**/*.ts",
+    "tsc": "tsc",
+    "build": "npm run tslint && npm run tsc",
+    "test": "npm run build",
```
princjef: i assume unit tests are coming in a future pr

princjef: actually i see the "unit" script below. any reason that's not part of `npm test`?

tsconfig.json
```diff
@@ -0,0 +1,40 @@
...
+    "noImplicitReturns": true,
+    "outDir": "dist",
+    "allowJs": false,
+    "noUnusedLocals":true,
+    "strict": true,
+    "strictPropertyInitialization": true,
+    "strictNullChecks": true,
+    "declaration": true,
+    "declarationDir": "./typings",
+    "lib": [
```
princjef: do you actually need all of these? if you're going to output es6 then i would expect this to be something like ["es6", "dom"] (assuming you actually need dom APIs, which you won't have in node)

package.json
```diff
@@ -0,0 +1,53 @@
...
+    "mocha": "^5.0.5",
+    "ts-node": "^5.0.1",
+    "tslint": "^5.9.1",
+    "typescript": "^2.9.1",
+    "dotenv": "^5.0.1",
+    "@types/dotenv": "^4.0.3",
+    "nyc": "^11.8.0"
+  },
+  "scripts": {
+    "tslint": "tslint -p . -c tslint.json --exclude examples/**/*.ts --exclude test/**/*.ts",
```
princjef: do you need the excludes here? looks like you specified excludes in your tslint.json, so i'd put them all in one place or the other

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: node_js
+
+node_js:
+  - 8
```
princjef: iirc we went back to node 6 with the event hubs package. any reason not to do so here as well? if we support it, we should include it in CI

package.json
```diff
@@ -0,0 +1,53 @@
...
+    "ms-rest-azure": "^2.5.5",
+    "rhea": "^0.2.14",
+    "tslib": "^1.9.2",
+    "uuid": "^3.2.1"
+  },
+  "devDependencies": {
+    "@types/uuid": "^3.4.3",
+    "@types/async-lock": "^1.1.0",
+    "@types/debug": "^0.0.30",
+    "@types/node": "^8.0.37",
```
princjef: same comment as the CI. if we want to support node 6 then use the node 6 typings so you don't accidentally use something that isn't actually available

tsconfig.json
```diff
@@ -0,0 +1,40 @@
...
+    "target": "es6",
+    "importHelpers": true,
+    "moduleResolution": "node",
+    "noImplicitReturns": true,
+    "outDir": "dist",
+    "allowJs": false,
+    "noUnusedLocals":true,
+    "strict": true,
+    "strictPropertyInitialization": true,
+    "strictNullChecks": true,
```
princjef: nit: this is already covered by "strict" (as are some other rules)

tsconfig.json
```diff
@@ -0,0 +1,40 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "preserveConstEnums": true,
+    "sourceMap": true,
+    "newLine": "LF",
+    "target": "es6",
```
princjef: if you want node 8 to be your min version, make this "es2017". you'll get native async/await and some other goodies to improve productivity/perf

.vscode/settings.json
amarzavery: The reason for keeping them here is: If someone wants to contribute to the project, then they don't have to worry about indentation for this project. It becomes easy to review PRs and improves productivity for the reviewer and the contributor.
The `.vscode/` dir is set in .`npmignore` so nothing from this dir is published.

package.json
amarzavery: More tests will be coming soon.
They are live tests and would need creds to be setup in Travis CI. Moreover, there is a danger of someone doing console.log() in their PR to get creds. Have to setup travis in a way that live tests are run only for merge commits. It is in the backlog.

tsconfig.json
amarzavery: One of the goals is to make this sdk isomorphic. Hence kept them for now. Can remove it as it is not needed at the moment.

package.json
amarzavery: I was thinking to restrict the min version of node.js to 8.x. What do you think?

examples/simpleReceive.ts
```diff
@@ -0,0 +1,51 @@
+import { OnMessage, OnError, MessagingError, delay, Message, ReceiveMode, Namespace } from "../lib";
+// import * as rhea from "rhea";
```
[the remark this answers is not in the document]
amarzavery: Nope. I was debugging something. Shall remove it.

examples/simpleReceive.ts
```diff
@@ -0,0 +1,51 @@
...
+const entityPath = "QUEUE_NAME";
+const str = process.env[connectionString] || "";
+const path = process.env[entityPath] || "";
+console.log("str: ", str);
+console.log("path: ", path);
+
+let ns: Namespace;
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+  const client = ns.createQueueClient(path, { receiveMode: ReceiveMode.peekLock, maxConcurrentCalls: 1 });
```
princjef: i know we talked about this, but can you remind me why the receive mode and concurrency are on the client rather than when you call `client.receive()`? they don't affect the sender at all so it seems odd for them to be at the client level

.vscode/settings.json
princjef: it makes sense for the indentation (though you might consider something like [editorconfig](https://editorconfig.org/) to support a broader set of tools). it's really the other settings. for instance, i actually like to see the node modules in the file tree but with these settings i don't have the choice because the repository itself has shut it off.

not a huge deal either way

package.json
princjef: ah yes forgot about that

tsconfig.json
princjef: if you want isomorphic then you can just exclude "lib" altogether. the default lib for any given target is the set of isomorphic definitions. then "dom" can be added back in if you need it (typically due to dependencies that pull it in)

package.json
princjef: that's fine by me. if people request node 6 we can always refactor later

examples/simpleReceive.ts
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      brokeredMessage.complete();
+    }
+  }
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+  //console.log(onMessage, onError);
+  client.receive(onMessage, onError, { autoComplete: true });
+  // const msgs = await client.receiveBatch(10);
```
princjef: probably worth cleaning up the commented out lines in here and maybe factoring the batch receive out into a separate example

examples/simpleSend.ts
```diff
@@ -0,0 +1,27 @@
...
+const path = process.env[entityPath] || "";
+console.log("str: ", str);
+console.log("path: ", path);
+
+let ns: Namespace;
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+  const client = ns.createQueueClient(path);
+  await client.send({ body: "Hello sb world!!" });
+  // await client.send({ body: "Hello awesome world!!2" + new Date().toString() });
```
princjef: same as the other example

lib/amqp-common/auth/aad.ts
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * @property {(ApplicationTokenCredentials | UserTokenCredentials | DeviceTokenCredentials | MSITokenCredentials)} credentials - The credentials object after successful authentication with AAD.
+   */
+  credentials: ApplicationTokenCredentials | UserTokenCredentials | DeviceTokenCredentials | MSITokenCredentials;
+  /**
+   * @property {number} tokenRenewalMarginInSeconds - The number of seconds within which it is
+   * good to renew the token. A constant set to 270 seconds (4.5 minutes). Adal has a set window of 5 minutes
+   * when it refreshes the token from its token cache.
+   */
+  readonly tokenRenewalMarginInSeconds: number = 270;
```
princjef: i'm probably misunderstanding but does this mean the token gets renewed every 4.5 minutes or every 55.5 minutes? an interval of 4.5 minutes when the token duration is 1 hour seems overzealous.

lib/amqp-common/auth/aad.ts
```diff
@@ -0,0 +1,71 @@
...
+    return new Promise<TokenInfo>((resolve, reject) => {
+      self.credentials.getToken((err, result) => {
+        if (err) {
+          reject(err);
+        }
+        let expiresOn = Date.now();
+        if (result.expiresOn && result.expiresOn instanceof Date) {
+          expiresOn = result.expiresOn.getTime();
+        }
+        const expiry = Math.floor(expiresOn / 1000) + self.tokenValidTimeInSeconds - 5;
```
princjef: nit: the `5` feels like it should be a constant somewhere

lib/message.ts
```diff
@@ -0,0 +1,682 @@
...
+    debug("SBMessage to AmqpMessage: %O", amqpMsg);
+    return amqpMsg;
+  }
+
+  export function fromAmqpMessage(msg: AmqpMessage): SBMessage {
+    if (!msg) {
+      throw new Error("msg cannot be null or undefined.");
+    }
+    const sbmsg: SBMessage = {
+      body: msg.body,
```
binzywu: >body [](start = 16, length = 4)

what is the type of this body? is it AMQP type or it is node.js type?

lib/client.ts
```diff
@@ -0,0 +1,52 @@
...
+  protected _context: ClientEntityContext;
+
+  /**
+   * Instantiates a client pointing to the ServiceBus entity given by this configuration.
+   *
+   * @constructor
+   * @param {string} name The entity name.
+   * @param {ConnectionContext} context The connection context to create the QueueClient.
+   * @param {TokenProvider} [tokenProvider] The token provider that provides the token for authentication.
+   * Default value: SasTokenProvider.
```
binzywu: is token provider will be added for future?

binzywu: is token provider will be added for future?

.travis.yml
amarzavery: was thinking of having min support to node.js 8.x since 10.x will become LTS this fall.

examples/simpleReceive.ts
amarzavery: I feel the same. 2 reasons for putting it at client level
- .net sdk has it at the client level.
- Wanted receiveBatch() and receive() both to have same modes. Moreover once receive() is called, it cannot be called another time without calling receiveHandler.stop(). Hence thought of having them as client level options.

examples/simpleReceive.ts
amarzavery: Oh yes. Definitely. Will do that.

lib/amqp-common/auth/aad.ts
amarzavery: Oh this is the margin (remaining time in 1 hour). This is used for AAD Token Credentials. If I try to renew the token before 55 minutes then ADAL will give me the stale token from its token cache. Hence I make sure that the call to get the token is made within adal's token renewal window, so I get a valid token for the next hour.

lib/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  parseConnectionString, ServiceBusDeliveryAnnotations, ServiceBusMessageAnnotations,
+  ServiceBusConnectionStringModel, delay, Timeout, MessagingError
+} from "./amqp-common";
+export {
+  AmqpError, Delivery, Dictionary, MessageProperties, MessageHeader
+} from "./rhea-promise";
+export { Message, ReceivedSBMessage, SBMessage } from "./message";
+export { ReceiveHandler } from "./streamingReceiver";
+export { ReceiveOptions, OnError, OnMessage } from "./messageReceiver";
+export { QueueClientOptions, QueueClient, ReceiveMode } from "./queueClient";
```
princjef: i think you're missing the `TopicClient` and `SubscriptionClient`

lib/client.ts
```diff
@@ -0,0 +1,52 @@
...
+  constructor(name: string, context: ConnectionContext) {
+    this.name = name;
+    this.id = `${name}/${uuid()}`;
+    this._context = ClientEntityContext.create(name, context);
+  }
+  /**
+   * Provides the current type of the Client.
+   * @return {string} The entity type.
+   */
+  get type(): string {
```
princjef: is this just for diagnostic purposes? does it need to be exposed to the user? (the fewer things that are exposed the fewer things you have to worry about maintaining backward compatibility for)

examples/simpleReceive.ts
princjef: i see. i can see it for receiveMode, but the only place where the maxConcurrentCalls has any meaning is on `receive()`. the batching receiver manages the credits in a different way so i think we should move `maxConcurrentCalls` down to the `receive()` options

lib/queueClient.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }
+
+  /**
+   * Sends the given message to the ServiceBus Queue.
+   *
+   * @param {any} data  Message to send.  Will be sent as UTF8-encoded JSON string.
+   * @returns {Promise<Delivery>} Promise<Delivery>
+   */
+  async send(data: SBMessage): Promise<Delivery> {
```
princjef: nit: i'd call it `ServiceBusMessage` for clarity

princjef: what's contained in a `Delivery`? it looks like the type comes all the way from the rhea library. i'm not sure what data is needed besides the error if something goes wrong. unless there's a strong need for it, i'd just return `Promise<void>` until you need something else

lib/queueClient.ts
```diff
@@ -0,0 +1,209 @@
...
+        // Close the sender.
+        if (this._context.sender) {
+          await this._context.sender.close();
+        }
+        // Close the receiver.
+        if (this._context.streamingReceiver) {
+          await this._context.streamingReceiver.close();
+        }
+        // Close the management session
+        await this._context.managementSession!.close();
```
princjef: is this management session shared between all clients on the namespace? if so, won't closing it break the other clients?

lib/queueClient.ts
```diff
@@ -0,0 +1,209 @@
...
+   * @param {OnMessage} onMessage          The message handler to receive Message objects.
+   * @param {OnError} onError              The error handler to receive an error that occurs
+   * while receiving messages.
+   *
+   * @returns {ReceiveHandler} ReceiveHandler - An object that provides a mechanism to stop
+   * receiving more messages.
+   */
+  receive(onMessage: OnMessage, onError: OnError, options?: MessageHandlerOptions): ReceiveHandler {
+    if (!this._context.streamingReceiver ||
+      (this._context.streamingReceiver && !this._context.streamingReceiver.isOpen())) {
```
princjef: the second check of `this._context.streamingReceiver` isn't needed. the full condition can be:

```ts
if (!this._context.streamingReceiver || !this._context.streamingReceiver.isOpen()) {
```

lib/queueClient.ts
```diff
@@ -0,0 +1,209 @@
...
+      let error: MessagingError | undefined;
+      let result: Message[] = [];
+      try {
+        result = await bReceiver.receive(maxMessageCount, maxWaitTimeInSeconds);
+      } catch (err) {
+        error = err;
+        debug("[%s] Receiver '%s', an error occurred while receiving %d messages for %d max time:\n %O",
+          this._context.namespace.connectionId, bReceiver.id, maxMessageCount, maxWaitTimeInSeconds, err);
+      }
+      if (error) {
```
princjef: why not just throw in the catch block rather than assigning to a variable and throwing immediately after?

lib/messageReceiver.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  constructor(context: ClientEntityContext, receiverType: ReceiverType, options?: ReceiveOptions) {
+    super(`${context.entityPath}`, context);
+    if (!options) options = {};
+    this.receiverType = receiverType;
+    this.address = `${this._context.entityPath}`;
+    this.audience = `${this._context.namespace.config.endpoint}${this._context.entityPath}`;
+    this.maxConcurrentCalls = options.maxConcurrentCalls != undefined ?
+      options.maxConcurrentCalls : 1;
+    this.autoComplete = options.autoComplete != undefined ? options.autoComplete : false;
```
princjef: nit: `this.autoComplete = !!options.autoComplete` is equivalent

lib/messageReceiver.ts
```diff
@@ -0,0 +1,232 @@
...
+    this.receiveMode = options.receiveMode || ReceiveMode.peekLock;
+    this._onAmqpMessage = async (context: EventContext) => {
+      const bMessage = new Message(context.message!, context.delivery!);
+      bMessage.body = this._context.namespace.dataTransformer.decode(context.message!.body);
+      try {
+        await this._onMessage!(bMessage);
```
princjef: having the `!` here is odd because this is depending on something that extends it filling in the implementation for it. if you want to make the assertion it probably makes more sense to do it on the declaration with `protected _onMessage!: OnMessage;`

princjef: i'd actually recommend having the `StreamingReceiver` and `BatchingReceiver` contain a `MessageReceiver` rather than extending it so you can just pass the necessary things in the constructor. that way you don't have to have these implicit contracts where one thing sets a property on the other

lib/messageReceiver.ts
```diff
@@ -0,0 +1,232 @@
...
+      this._onError!(sbError);
+    };
+  }
+
+  /**
+   * Determines whether the AMQP receiver link is open. If open then returns true else returns false.
+   * @returns {boolean} boolean
+   */
+  isOpen(): boolean {
+    return this._receiver! && this._receiver!.isOpen();
```
princjef: nit: the `!` undermines the check you're making from a type perspective. just do `this._receiver && this._receiver.isOpen()` the first check is enough for typescript to infer that `this._receiver` exists for the second one

lib/linkEntity.ts
```diff
@@ -0,0 +1,164 @@
...
+   * @return {Promise<void>} Promise<void>
+   */
+  protected async _negotiateClaim(setTokenRenewal?: boolean): Promise<void> {
+    // Acquire the lock and establish a cbs session if it does not exist on the connection.
+    // Although node.js is single threaded, we need a locking mechanism to ensure that a
+    // race condition does not happen while creating a shared resource (in this case the
+    // cbs session, since we want to have exactly 1 cbs session per connection).
+    debug("[%s] Acquiring cbs lock: '%s' for creating the cbs session while creating the %s: " +
+      "'%s' with address: '%s'.", this._context.namespace.connectionId,
+      this._context.namespace.cbsSession!.cbsLock,
```
princjef: what guarantess the existence of the cbsSession?

lib/managementClient.ts
```diff
@@ -0,0 +1,134 @@
...
+  address?: string;
+  audience?: string;
+}
+
+/**
+ * @class ManagementClient
+ * Descibes the EventHubs Management Client that talks
+ * to the $management endpoint over AMQP connection.
+ */
+export class ManagementClient extends LinkEntity {
```
princjef: doesn't look like you're actually using this anywhere yet. is that accurate? assuming this is future planning for renew lock and such

vIceBerg:
Hi!

How is it going? Is there any chance to see this project available soon?

amarzavery:
@vIceBerg - We published a GA version of the event hubs sdk based on amqp protocol yesterday. Now the focus will be on ServiceBus.

.travis.yml
amarzavery: Will be supporting 6.x onwards. Have updated the readme and target field in tsconfig.json as well.

lib/queueClient.ts
amarzavery: People might need Delivery for tags associated with it. Hence keeping it over there.

lib/messageReceiver.ts
amarzavery: Without the excalamation mark, I see the following error
> Type 'boolean | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.

Transforming to this `return this._receiver ? this._receiver.isOpen() : false;` makes tsc happy.

lib/messageSender.ts
```diff
@@ -0,0 +1,523 @@
...
+            `to operation timeout.`;
+          log.error(desc);
+          const e: AmqpError = {
+            condition: ErrorNameConditionMapper.ServiceUnavailableError,
+            description: desc
+          };
+          return reject(translate(e));
+        };
+
+        this._sender!.on(SenderEvents.accepted, onAccepted);
```
princjef: these will still run up against the event emitter warnings for listener count right?

package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@azure/service-bus",
+  "version": "0.1.0",
+  "description": "Azure ServiceBus SDK for JS",
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "main": "./dist/lib/index.js",
+  "types": "./typings/lib/index.d.ts",
+  "dependencies": {
+    "@azure/amqp-common": "azure/amqp-common-js",
```
princjef: is this intentional?

lib/messageSender.ts
amarzavery: Yup. I have set the max limit to 1000 for now.
```
this._sender.setMaxListeners(1000);
```

package.json
amarzavery: I have to make changes to the EventProcessorHost package, before publishing the 0.1.4 version of amqp-common. Hence, have taken a dependency on the git repo. Will update this after making changes to EPH sdk.

PR #6, opened 2018-10-18
Add GetQueue specification

ghost:
Add management plane swagger specification that defines the GetQueue operation.
I also added tests that verify/document the behavior of the operation.


.gitignore
```diff
@@ -51,9 +51,12 @@ packages
 *.user
 .ntvs_analysis.dat
 obj/*
-.env
+*.env
```
amarzavery: I wanted to have sample.env in the repo. It shows customers the environment variables that they need to use. It acts as a template for them. Hence just .env would be good.

dotnet/ServiceBusDotNetTests.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
amarzavery: Please add `dotnet` in .npmignore to ensure we do not publish this by accident.

lib/management/models/index.ts
```diff
@@ -0,0 +1,350 @@
+/*
```
amarzavery: Do you want to put the generated stuff under `lib/management/generated`. In that way if we want to write custom stuff it will not get mixed with generated stuff.

lib/management/models/index.ts
```diff
@@ -0,0 +1,350 @@
...
+  accessedAt?: string;
+  /**
+   * @member {boolean} [supportOrdering] TODO: Add description
+   */
+  supportOrdering?: boolean;
+  /**
+   * @member {GetQueueResponseContentQueueDescriptionCountDetails}
+   * [countDetails] TODO: Add description
+   */
+  countDetails?: GetQueueResponseContentQueueDescriptionCountDetails;
```
amarzavery: Is it possible to shorten the type name to `QueueDescriptionCountDetails`?

specification/examples/CreateQueue.example.xml
```diff
@@ -0,0 +1,67 @@
+PUT https://daschulttest1.servicebus.windows.net/testQueuePath?api-version=2017-04 HTTP/1.1
+Authorization: SharedAccessSignature sr=https%3A%2F%2Fdaschulttest1.servicebus.windows.net%2FtestQueuePath%2F&sig=bCEWzr1DOgXB5XCeNjWJU2zfhH1%2F%2Fca56gVNovbPPUY%3D&se=1539812750&skn=RootManageSharedAccessKey
```
amarzavery: Is this the real SharedAccessKey? If so please delete this key from the portal, remove it from here and make sure to either not have this header or fill it with dummy value.

test/serviceBusManagementClientTests.ts
```diff
@@ -0,0 +1,241 @@
+import { SasTokenProvider, TokenInfo } from "@azure/amqp-common";
```
amarzavery: Can you suffix the test files with spec.ts. I learnt about this new convention and like it.

tsconfig.json
```diff
@@ -23,6 +23,7 @@
   ],
   "include": [
     "./lib/**/*.ts",
-    "./examples/**/*.ts"
+    "./examples/**/*.ts",
+    "./test/**/*.ts"
```
amarzavery: Please make sure to add `dist/test` and `typings/test` to .npmignore

tslint.json
```diff
@@ -8,7 +8,6 @@
     ],
     "typedef": [
       true,
-      "call-signature",
```
amarzavery: any reason to remove these from tsconfig?

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+        "description": "The name of the ServiceBus namespace to send requests to.",
+        "required": true,
+        "type": "string",
+        "in": "path",
+        "x-ms-skip-url-encoding": true
+      }
+    ]
+  },
+  "paths": {
+    "/{queuePath}/": {
```
amarzavery: Do we need to have forward slash in the end?

amarzavery (changes requested):

.gitignore
daschult: Done.

dotnet/ServiceBusDotNetTests.sln
daschult: Done.

lib/management/models/index.ts
daschult: Done.

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+      "get": {
+        "tags": [
+          "Queues"
+        ],
+        "operationId": "Queue_Get",
```
amarzavery: Please make sure that the entry in the tags array matches the initial part in the operationId. The operationId should also be "Queues_Get". We recommend having plural nouns for the operation group.

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+          },
+          {
+            "$ref": "#/parameters/EnrichFalseParameter"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK -- Get Queue operation completed successfully.",
+            "schema": {
+              "$ref": "#/definitions/GetQueueResponse"
```
amarzavery: Let's name this `Queue`. This will ensure that the generated interface is called Queue.

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+            }
+          }
+        }
+      },
+      "put": {
+        "tags": [
+          "Queues"
+        ],
+        "operationId": "Queue_Create",
```
amarzavery: Queues_Create

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+            "href": {
+              "description": "TODO: Add description",
+              "type": "string",
+              "xml": {
+                "attribute": true
+              }
+            }
+          }
+        },
+        "content": {
```
amarzavery: This should be a top level model in "definitions" and content should have a reference to that model

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+          "type": "object",
+          "properties": {
+            "type": {
+              "description": "TODO: Add description",
+              "type": "string",
+              "xml": {
+                "attribute": true
+              }
+            },
+            "QueueDescription": {
```
amarzavery: This should be a top level model in the definitions object.

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+                },
+                "MessageCount": {
+                  "description": "TODO: Add description",
+                  "type": "integer"
+                },
+                "IsAnonymousAccessible": {
+                  "description": "TODO: Add description",
+                  "type": "boolean"
+                },
+                "Status": {
```
amarzavery: Some of the properties from here onwards look like readOnly candidates

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+                "AccessedAt": {
+                  "description": "TODO: Add description",
+                  "type": "string"
+                },
+                "SupportOrdering": {
+                  "description": "TODO: Add description",
+                  "type": "boolean"
+                },
+                "CountDetails": {
+                  "description": "TODO: Add description",
```
amarzavery: This should be a top level model in the definitions object.

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+        "name": "entry",
+        "namespace": "http://www.w3.org/2005/Atom"
+      }
+    },
+    "CreateQueueBody": {
+      "description": "The response from a CreateQueue operation.",
+      "type": "object",
+      "properties": {
+        "content": {
+          "description": "TODO: Add description",
```
amarzavery: This should be a top level model in the definitions object.

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+        "author",
+        "link",
+        "content"
+      ],
+      "xml": {
+        "name": "entry",
+        "namespace": "http://www.w3.org/2005/Atom"
+      }
+    },
+    "CreateQueueBody": {
```
amarzavery: `CreateQueueParameters` would be a better name

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+      "description": "The response from a CreateQueue operation.",
```
amarzavery: Parameters that can be provided while creating a queue.

specification/ServiceBusManagement.json
```diff
@@ -0,0 +1,413 @@
...
+                },
+                "EnableBatchedOperations": {
+                  "description": "TODO: Add description",
+                  "type": "boolean",
+                  "default": true
+                },
+                "Status": {
+                  "description": "TODO: Add description",
+                  "type": "string",
+                  "default": "Active"
```
amarzavery: Wouldn't this be readonly? If not "Active" then what are the other possible values that one can provide? Can someone create an "InActive" queue?

amarzavery (changes requested):

PR #8, opened 2018-10-30
Implement several $management operations

amarzavery:


examples/msgRenewLock.ts
```diff
@@ -0,0 +1,40 @@
...
+console.log("path: ", path);
+
+let ns: Namespace;
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+  const client = ns.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(">>> Message: ", brokeredMessage);
+    console.log("### Actual message:", brokeredMessage.body ? brokeredMessage.body.toString() : null);
+    const result = await client.renewLock(brokeredMessage.lockToken!);
```
princjef: does it make sense for the user to actually provide the lock token or to just provide the message itself? that way they wouldn't have to worry about the implementation details of how locking works

princjef: it also allows you to set the `lockedUntilUtc` for them rather than them setting it outside of the call

lib/managementClient.ts
```diff
@@ -77,37 +129,401 @@ export class ManagementClient extends LinkEntity {
...
+
+    if (typeof receiveMode !== 'number') {
+      throw new Error("'receiveMode' is a required parameter with value 1 or 2.");
+    }
+
+    const messageList: ServiceBusMessage[] = [];
+
+    try {
+      const messageBody: any = {};
+      messageBody["sequence-numbers"] = types.wrap_array(sequenceNumbers.map((i) => { return Buffer.from(i.toBytesBE()); }), 0x81, undefined);
```
princjef: nit: you can use the arrow function shorthand here

```suggestion
      messageBody["sequence-numbers"] = types.wrap_array(sequenceNumbers.map((i) => Buffer.from(i.toBytesBE())), 0x81, undefined);
```

There's also a [tslint rule](https://palantir.github.io/tslint/rules/arrow-return-shorthand/) if you want it automatically enforced

lib/queueClient.ts
```diff
@@ -165,4 +200,101 @@ export class QueueClient extends Client {
...
+  /**
+   * Receives a specific deferred message identified by `sequenceNumber` of the `Message`.
+   * @param sequenceNumber The sequence number of the message that will be received.
+   * @returns Promise<ServiceBusMessage | undefined>
+   * - Returns `Message` identified by sequence number.
+   * - Returns `undefined` if no such message is found.
+   * - Throws an error if the message has not been deferred.
+   */
+  async receiveDeferredMessage(sequenceNumber: Long): Promise<ServiceBusMessage | undefined> {
+    if (this.receiveMode !== ReceiveMode.peekLock) {
```
princjef: It feels weird to be enforcing the receive mode here. These operations don't really use the receive mode in the way that the normal messaging operations do, so it feels like people should be allowed to use them regardless of configured mode.

examples/msgRenewLock.ts
princjef: Nevermind i see what you're doing. I'd probably have the example show passing the full message because i think that's likely to be the more common pattern

lib/queueClient.ts
amarzavery: I see, what you are saying. Receivemode on the client has no bearing on receiving deferred messages from the $management endpoint. Disposing the message happens via updateDispositionStatus() on the $management endpoint. However, the updateDispositionStatus() happens within, defer(), complete(), abandon(), deadLetter() methods. The customer does not have to explicitly call updateDispositionStatus().

Now if the receiveMode is (let's say) ReceiveAndDelete, wont it be a little confusing for customers, if they are able to dispose off messages using message.defer|complete|abandon|deadletter()? 


I saw the same behavior in .net sdk. It [throws an error, if the receive mode is not peek lock](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs#L434). Hence, replicated the behavior over here for consistency.

princjef: Ah okay. Hadn't dealt with the `updateDispositionStatus()` stuff before. If that's the usage then I think it makes sense to constrain that one to avoid weird errors

princjef: Didn't see the comment about the .NET SDK before. I'm alright with keeping things this way in the name of consistency.

examples/msgRenewLock.ts
amarzavery: The method takes a union of lockToken | Message. Yup, shall update the example.

PR #9, opened 2018-10-31
autorenew lock for messages received by streaming receiver

amarzavery:
- add relevant methods to subscription and topic client and update documentation of some of the methods
- autorenew lock for messages received by streaming receiver

### DIFF
--- a/examples/receiveAutoRenewLock.ts
+++ b/examples/receiveAutoRenewLock.ts
@@ -1,0 +1,40 @@
+import { OnMessage, OnError, MessagingError, delay, ServiceBusMessage, ReceiveMode, Namespace } from "../lib";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+const connectionString = "SERVICEBUS_CONNECTION_STRING";
+const entityPath = "QUEUE_NAME";
+const str = process.env[connectionString] || "";
+const path = process.env[entityPath] || "";
+console.log("str: ", str);
+console.log("path: ", path);
+
+let ns: Namespace;
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+  const client = ns.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
+  // Please note: Lock duration property on the Queue was set to 45 seconds.
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(">>> Message: ", brokeredMessage);
+    console.log("### Actual message:", brokeredMessage.body ? brokeredMessage.body.toString() : null);
+    const time = 30000;
+    console.log(">>>> Sleeping for %d seconds. Meanwhile autorenew of message lock should happen.", time / 1000);
+    await delay(time);
+    console.log(">>>> Sleeping for %d seconds again. Meanwhile autorenew of message lock should happen.", time / 1000);
+    await delay(time);
+    console.log(">>>> Now finally exiting from the message handler code...");
+  }
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+  const rcvHandler = client.receive(onMessage, onError, { maxAutoRenewDurationInSeconds: 350 });
+  await delay(30000000);
+  await rcvHandler.stop();
+}
+
+main().then(() => {
+  console.log(">>>> Calling close....");
+  return ns.close();
+}).catch((err) => {
+  console.log("error: ", err);
+});
--- a/examples/receiveWithoutAutoLockRenewAndComplete.ts
+++ b/examples/receiveWithoutAutoLockRenewAndComplete.ts
@@ -1,0 +1,50 @@
+import {
+  OnMessage, OnError, MessagingError, delay, ServiceBusMessage, ReceiveMode, Namespace,
+  MessageHandlerOptions
+} from "../lib";
+import * as dotenv from "dotenv";
+import { } from '../lib/streamingReceiver';
+dotenv.config();
+
+const connectionString = "SERVICEBUS_CONNECTION_STRING";
+const entityPath = "QUEUE_NAME";
+const str = process.env[connectionString] || "";
+const path = process.env[entityPath] || "";
+console.log("str: ", str);
+console.log("path: ", path);
+
+let ns: Namespace;
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+  const client = ns.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
+  // Please note: Lock duration property on the Queue was set to 45 seconds.
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(">>> Message: ", brokeredMessage);
+    console.log("### Actual message:", brokeredMessage.body ? brokeredMessage.body.toString() : null);
+    const time = 50000;
+    console.log(">>>> Sleeping for %d seconds. Meanwhile autorenew of message lock should NOT happen.", time / 1000);
+    await delay(time);
+    try {
+      await brokeredMessage.complete();
+    } catch (err) {
+      console.log("This should error, since the lock would have expired: %o.", err);
+    }
+  }
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+  const handlerOptions: MessageHandlerOptions = {
+    autoComplete: false,
+    maxAutoRenewDurationInSeconds: 0
+  };
+  const rcvHandler = client.receive(onMessage, onError, handlerOptions);
+  await delay(30000000);
+  await rcvHandler.stop();
+}
+
+main().then(() => {
+  console.log(">>>> Calling close....");
+  return ns.close();
+}).catch((err) => {
+  console.log("error: ", err);
+});
--- a/lib/messageReceiver.ts
+++ b/lib/messageReceiver.ts
@@ -5,11 +5,13 @@ import * as log from "./log";
 import { LinkEntity } from "./linkEntity";
 import { ClientEntityContext } from "./clientEntityContext";
 import {
-  translate, Constants, MessagingError, retry, RetryOperationType, RetryConfig
+  translate, Constants, MessagingError, retry, RetryOperationType, RetryConfig,
+  ConditionErrorNameMapper
 } from "@azure/amqp-common";
 import { Receiver, OnAmqpEvent, EventContext, ReceiverOptions, AmqpError } from "rhea-promise";
 import { ServiceBusMessage } from "./serviceBusMessage";
-import { getUniqueName } from "./util/utils";
+import { getUniqueName, calculateRenewAfterDuration } from "./util/utils";
+import { MessageHandlerOptions } from "./streamingReceiver";
 
 /**
  * @interface CreateReceiverOptions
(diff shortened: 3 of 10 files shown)


examples/receiveAutoRenewLock.ts
```diff
@@ -0,0 +1,40 @@
...
+    console.log(">>>> Sleeping for %d seconds. Meanwhile autorenew of message lock should happen.", time / 1000);
+    await delay(time);
+    console.log(">>>> Sleeping for %d seconds again. Meanwhile autorenew of message lock should happen.", time / 1000);
+    await delay(time);
+    console.log(">>>> Now finally exiting from the message handler code...");
+  }
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+  const rcvHandler = client.receive(onMessage, onError, { maxAutoRenewDurationInSeconds: 350 });
```
amarzavery: default value is 300.

examples/receiveWithoutAutoLockRenewAndComplete.ts
```diff
@@ -0,0 +1,50 @@
...
+    } catch (err) {
+      console.log("This should error, since the lock would have expired: %o.", err);
+    }
+  }
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+  const handlerOptions: MessageHandlerOptions = {
+    autoComplete: false,
+    maxAutoRenewDurationInSeconds: 0
```
amarzavery: by setting it 0, autolock renewal is disabled.

examples/receiveWithoutAutoLockRenewAndComplete.ts
```diff
@@ -0,0 +1,50 @@
...
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(">>> Message: ", brokeredMessage);
+    console.log("### Actual message:", brokeredMessage.body ? brokeredMessage.body.toString() : null);
+    const time = 50000;
+    console.log(">>>> Sleeping for %d seconds. Meanwhile autorenew of message lock should NOT happen.", time / 1000);
+    await delay(time);
+    try {
+      await brokeredMessage.complete();
+    } catch (err) {
+      console.log("This should error, since the lock would have expired: %o.", err);
```
amarzavery: I was expecting this scenario to throw an error. Servicebus sends the error as seen on line 6 in the debug logs. However, I am not sure how to intercept it.

Looks like I will have to listen for `"settle"` event on the receiver link and resolve/reject the promise based on some information.
Question 1: Not sure where I can find that disposition information in the EventContext for "settle" event in the receiver?
Question 2: When sending disposition for multiple messages, how would I know which disposition was received for which message?

```
  1. rhea:frames [connection-1]:1 -> disposition#15 {"role":true,"state":[]}  +5s
  2. rhea:raw [connection-1] SENT: 28 0000001c02000001005315d00000000c000000054143434200532445 +5s
  3. rhea:io [connection-1] read 366 bytes +5s
  4. rhea:io [connection-1] got frame of size 366 +0ms
  5. rhea:raw [connection-1] RECV: 366 0000016e02000001005315d00000015e0000000642434041005325d00000014d0000000100531dd00000014100000003a31f636f6d2e6d6963726f736f66743a6d6573736167652d6c6f636b2d6c6f7374b100000116546865206c6f636b20737570706c69656420697320696e76616c69642e2045697468657220746865206c6f636b20657870697265642c206f7220746865206d6573736167652068617320616c7265616479206265656e2072656d6f7665642066726f6d207468652071756575652e205265666572656e63653a38383565323561352d646237612d343936312d623432392d6463346536346265396162312c20547261636b696e6749643a39626136306331353030303030303036303034623965326535626439323635645f47385f4232382c2053797374656d547261636b65723a6e6f646573646b3a71756575653a74717e3131312c2054696d657374616d703a31302f33312f3230313820333a35303a343020414d4040 +412ms
  6. rhea:frames [connection-1]:1 <- disposition#15 {"settled":true,"state":[["com.microsoft:message-lock-lost","The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue. Reference:885e25a5-db7a-4961-b429-dc4e64be9ab1, TrackingId:9ba60c1500000006004b9e2e5bd9265d_G8_B28, SystemTracker:nodesdk:queue:tq~111, Timestamp:10/31/2018 3:50:40 AM",null]]}  +413ms
  7. rhea:events [connection-1] Received disposition for incoming transfers +5s
  8. rhea:events [connection-1] Link got event: settled +0ms
  9. rhea-promise:receiver [connection-1] receiver got event: 'settled'. Re-emitting the translated context. +5s
  10. rhea-promise:translate [connection-1] Translating the context for event: 'settled'. +5s
```

examples/receiveWithoutAutoLockRenewAndComplete.ts
```diff
@@ -0,0 +1,50 @@
+import {
+  OnMessage, OnError, MessagingError, delay, ServiceBusMessage, ReceiveMode, Namespace,
+  MessageHandlerOptions
+} from "../lib";
+import * as dotenv from "dotenv";
+import { } from '../lib/streamingReceiver';
```
princjef: remove?

examples/receiveWithoutAutoLockRenewAndComplete.ts
princjef: might be worth adding a comment in the example itself explaining that

lib/messageReceiver.ts
```diff
@@ -163,29 +164,99 @@ export class MessageReceiver extends LinkEntity {
...
         context.delivery!);
+      let timer: NodeJS.Timer | undefined = undefined;
+      let continueExecution: boolean = true;
+      if (this.autoRenewLock) {
+        // - We need to renew locks before they expire by looking at bMessage.lockedUntilUtc.
+        // - This autorenewal needs to happen **NO MORE** than maxAutoRenewDurationInSeconds
+        // - We should be able to clear the renewal timer when the user's message handler
+        // is done (whether it succeeds or fails).
+        log.receiver("[%s] message with id '%s' is locked until %s.",
+          connectionId, bMessage.messageId, bMessage.lockedUntilUtc!.toString());
```
princjef: Is `lockedUntilUtc` set by Service Bus itself? If yes, then you can run into problems with clock skew between the client and the server. I don't remember what information is available, but i would base your timing on the duration of the lock rather than actual timestamps, particularly if the timestamp is generated by Service Bus. If the timestamp is generated by you this isn't as big of a deal (though still potentially a problem in certain situations).

lib/messageReceiver.ts
```diff
@@ -163,29 +164,99 @@ export class MessageReceiver extends LinkEntity {
...
+                log.receiver("[%s] Successfully renewed the lock for message with id '%s'.",
+                  connectionId, bMessage.messageId);
+              } catch (err) {
+                log.error("[%s] An error occured while auto renewing the message lock '%s' for " +
+                  "message with id '%s': %O.", connectionId, bMessage.lockToken,
+                  bMessage.messageId, err);
+              }
+              log.receiver("[%s] Calling the autorenewlock task again for message with id '%s'.",
+                connectionId, bMessage.messageId);
+              autoRenewLockTask();
```
princjef: Why run this in the case of a failed lock renewal? If you're renewing at 75-80% of the total lock duration and then fail and wait another 75-80%, your lock will be long expired by the time things come back around again. You might consider building in some faster retries for specific types of transient failures, but if those fail, then the auto-renewal seems pretty much dead in the water for that message. I'd move the log and requeue code into the `try` branch of this function.

Another thing to be aware of in the future: you can renew locks for multiple messages in a single call. It's more complicated to manage coalescing the renewals, but you'll reduce the chattiness between the client and server, which may be worth it for high volume scenarios. I wouldn't worry about it now, but it's a good optimization to be aware of should you need it down the line.

lib/messageReceiver.ts
```diff
@@ -163,29 +164,99 @@ export class MessageReceiver extends LinkEntity {
...
+        continueExecution = false;
+        clearTimeout(timer);
+        log.receiver("[%s] Stopping lock renewal for message with id '%s'.", connectionId,
+          bMessage.messageId);
       } catch (err) {
-        log.error("[%s] Abandoning the message with id '%s' on the receiver '%s' since an error " +
-          "occured: %O.", connectionId, bMessage.messageId, this.name);
+        // Do not want renewLock to happen unnecessarily, while abandoning the message. Hence,
+        // doing this here. Otherwise, this should be done in finally.
+        continueExecution = false;
```
princjef: This is probably DRY overkill, but you could potentially add an inline function before the `try`/`catch` that you call in both places which will clear the timer and set `continueExecution` to false. With only two lines may not be worth it now, but I'd consider it if you start having more logic shared between the two

lib/messageReceiver.ts
```diff
@@ -163,29 +164,99 @@ export class MessageReceiver extends LinkEntity {
...
+    this.receiveMode = options.receiveMode || ReceiveMode.peekLock;
     this.maxConcurrentCalls = options.maxConcurrentCalls != undefined ?
       options.maxConcurrentCalls : 1;
     this.autoComplete = !!options.autoComplete;
-    this.receiveMode = options.receiveMode || ReceiveMode.peekLock;
+    this.maxAutoRenewDurationInSeconds = options.maxAutoRenewDurationInSeconds != undefined
+      ? options.maxAutoRenewDurationInSeconds
+      : 300;
+    this.autoRenewLock = this.maxAutoRenewDurationInSeconds > 0 && this.receiveMode === ReceiveMode.peekLock;
     this._onAmqpMessage = async (context: EventContext) => {
```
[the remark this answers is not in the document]
amarzavery: Get the essence of it. However, the problem is that the type of the handler is
```js
export declare type OnAmqpEvent = (context: EventContext) => void;
```
If I turn this into 
```js
export declare type OnAmqpEvent = (context: EventContext) => Promise<void>;
```
let's say just in rhea-promise. Will this still work well with rhea?

lib/managementClient.ts
```diff
@@ -213,16 +213,20 @@ export class ManagementClient extends LinkEntity {
...
       throw new Error("'lockTokenOrMessage' is a required parameter.");
     }
     if (typeof lockTokenOrMessage !== "object" && typeof lockTokenOrMessage !== "string") {
       throw new Error("'lockTokenOrMessage must be of type 'string' or of type 'object'.");
     }
-
+    if (!options) options = {};
+    if (options.delayInSeconds == undefined) options.delayInSeconds = 1;
+    if (options.timeoutInSeconds == undefined) options.timeoutInSeconds = 5;
+    if (options.times == undefined) options.times = 5;
```
amarzavery: default values were 15, 10, 3 which were pretty long. Hence customized it.

lib/messageReceiver.ts
```diff
@@ -528,12 +532,31 @@ export class MessageReceiver extends LinkEntity {
     }
   }
 
+  /**
+   * Disposes the message with the specified disposition.
+   * @param delivery Delivery associated with the message.
+   * @param operation The disposition type.
+   * @param [options] optional parameters that can be provided while disposing the message.
+   */
   async disposeMessage(delivery: Delivery, operation: DispositionType, options?: DispositionOptions): Promise<any> {
```
princjef: Didn't notice this before, but `disposeMessage` is a bit of a weird name. "dispose" and "disposition" don't really mean the same thing. I'd probably call this `settleMessage` or something like that

PR #16, opened 2018-11-16
Initial implementation of Sessions

amarzavery:


lib/session/sessionManager.ts
```diff
@@ -0,0 +1,75 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { OnSessionMessage, SessionHandlerOptions, MessageSession } from "./messageSession";
+import { OnError } from "../core/messageReceiver";
+import { ClientEntityContext } from "../clientEntityContext";
+import { getProcessorCount } from "../util/utils";
+import * as log from "../log";
+
+export class SessionManager {
+  /**
```
amarzavery: Have added very basic stuff. This needs more changes to ensure things work as expected.

amarzavery:
@AlexGhiondea , @ramya-rao-a  - Please review.

lib/clientEntityContext.ts
```diff
@@ -44,6 +51,11 @@ export interface ClientEntityContextBase {
    * client entity for receiving a batch of messages.
    */
   batchingReceiver?: BatchingReceiver;
+  /**
+   * @property {Dictionary<MessageSession>} messageSessions A dictionary of the MessageSession
+   * objects associated with this client.
+   */
+  messageSessions: Dictionary<MessageSession>;
```
ramya-rao-a: Why not use the built-in [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of `Dictionary`? Same question on all other usages of `Dictionary`?

lib/serviceBusMessage.ts
```diff
@@ -198,23 +198,23 @@ export module SendableMessageInfo {
       throw new Error("'msg' cannot be null or undefined.");
     }
 
-    if (msg.contentType && typeof msg.contentType !== "string") {
+    if (msg.contentType != undefined && typeof msg.contentType !== "string") {
```
ramya-rao-a: Why the move from `if (a)` to `if (a != undefined)`? Same for other checks in this file that are updated in this file. Are we trying to catch a particular case here?

lib/util/constants.ts
```diff
@@ -6,3 +6,4 @@ export const packageJsonInfo = {
   version: "0.1.0"
 };
 export const messageDispositionTimeout = 20000;
+export const defaultLastUpdatedTime = 259200000; // 3 * 24 * 3600 * 1000
```
ramya-rao-a: By its name, its not clear what this constant is going to be used for...

lib/util/utils.ts
```diff
@@ -1,8 +1,21 @@
...
 
+import * as Long from "long";
 import * as log from "../log";
 import { generate_uuid, string_to_uuid } from "rhea-promise";
+import { isBuffer } from "util";
+
+/**
+ * A constant that indicates whether the environment is node.js or browser based.
+ */
+export const isNode = typeof navigator === "undefined" && typeof process !== "undefined";
```
ramya-rao-a: Have you see this pattern elsewhere?
I am wondering if this is the best way to decide `isNode`

lib/session/sessionManager.ts
```diff
@@ -0,0 +1,75 @@
...
+    return this._maxConcurrentSessions;
+  }
+  private _maxConcurrentSessions: number;
+  /**
+   * @property {number} _maxConcurrentAcceptSessionRequests The maximum number of acceptSession
+   * requests that can be made concurrently at any given time.
+   */
+  private set _maxConcurrentAcceptSessionRequests(value: number) {
+
+    this._maxConcurrentAcceptSessionRequests = Math.min(value, getProcessorCount());
```
ramya-rao-a: Wont this result in an infinite setter because the name `_maxConcurrentAcceptSessionRequests` is used both for the setter and what is being set inside?

lib/session/sessionManager.ts
```diff
@@ -0,0 +1,75 @@
...
+   * sessions.
+   * - **Default**: `false`.
+   */
+  isManagingSessions: boolean = false;
+  /**
+   * @property {number} maxConcurrentSessions The maximum number of sessions that the user wants to
+   * handle concurrently.
+   * - **Default**: `2000`.
+   */
+  set maxConcurrentSessions(value: number) {
```
ramya-rao-a: `maxConcurrentSessions` only seems to be used to set `_maxConcurrentAcceptSessionRequests`.
In which case, why even track `maxConcurrentSessions` as a public property?

lib/queueClient.ts
```diff
@@ -309,4 +331,44 @@ export class QueueClient extends Client {
...
+   */
+  async listMessageSessions(skip: number, top: number, lastUpdatedTime?: Date): Promise<ListSessionsResponse> {
+    return this._context.managementClient!.listMessageSessions(skip, top, lastUpdatedTime);
+  }
+
+  /**
+   * 
+   * @param options 
+   */
+  async acceptSession(options?: AcceptSessionOptions): Promise<MessageSession> {
```
ramya-rao-a: From the code I see that this creates a receiver for given sessionId/receiver name.
But the name `acceptSession` doesn't convey the same. Is this what other SDKs call it as well?

lib/queueClient.ts
```diff
@@ -309,4 +331,44 @@ export class QueueClient extends Client {
...
+  }
+
+  /**
+   * Receives messages from a session enabled Queue.
+   * @param onSessionMessage The message handler to receive service bus messages from a session
+   * enabled Queue.
+   * @param onError The error handler to receive an error that occurs while receiving messages
+   * from a session enabled Queue.
+   */
+  receiveMessgesFromSessions(
```
ramya-rao-a: I am guessing this is analogous to the `receive`?
Can the user call `receiveMessgesFromSessions` without calling `acceptSession`?

lib/clientEntityContext.ts
amarzavery: Yup we could. Started using a dictionary/JS Object in EventHubs and then looked at Maps while implementing Event Processor Host. Hence started using Maps in some places, since then. 

Was wondering if there should be a preference for one over another. 

Looks like both have their own use cases.
[This article](https://medium.com/front-end-hacking/es6-map-vs-object-what-and-when-b80621932373) provides good info. This [SO link](https://stackoverflow.com/questions/44321324/javascript-map-much-slower-than-object-for-random-look-ups) also provides good info on Maps being slower than objects for lookups.

lib/serviceBusMessage.ts
amarzavery: Empty strings are allowed which the previous checks will not allow. Hence updating the code to specifically check for null and undefined.

lib/session/sessionManager.ts
amarzavery: Not really. I still need to add code to SessionManager. Which will make this a bit clearer. As mentioned above SessionManager code as seen until now is just a placeholder.

lib/session/sessionManager.ts
amarzavery: That is true. I am aware of it. Will be updating this file in a day or two. Have been working on SessionManager locally. Haven't updated those changes yet.

lib/util/constants.ts
amarzavery: It is used to provide the latUpdatedTime parameter while enumerating sessions. It is a required parameter. However, both Anna and me thought that it is better to set some default value for that parameter to provide a better developer experience and make this an optional parameter.

Would `defaultLastUpdatedTimeForListingSessions` be a better name?

lib/util/utils.ts
amarzavery: Yup. We have exported this function in `ms-rest-js` (runtime for Autorest generated TS libraries). It is a well tested method in context of Autorest generated libraries. Btw, found similar approaches on SO to detect whether one is in the browser or node.js environment.

lib/core/managementClient.ts
```diff
@@ -532,6 +615,190 @@ export class ManagementClient extends LinkEntity {
...
+   * @param sessionId The session for which the state needs to be set
+   * @param state The state that needs to be set.
+   * @returns Promise<void>
+   */
+  async setSessionState(sessionId: string, state: any): Promise<void> {
+    if (typeof sessionId !== "string") {
+      throw new Error("'sessionId' is a required parameter and must be of type 'string'.");
+    }
+
+    if (state == undefined) {
```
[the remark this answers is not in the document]
amarzavery: That is true. Will change that.

lib/core/managementClient.ts
```diff
@@ -532,6 +615,190 @@ export class ManagementClient extends LinkEntity {
...
+  }
+
+  /**
+   * Lists the sessions on the ServiceBus Queue/Topic.
+   * @param lastUpdateTime Filter to include only sessions updated after a given time.
+   * @param skip The number of sessions to skip
+   * @param top Maximum numer of sessions.
+   * @returns Promise<ListSessionsResponse> A list of sessions.
+   */
+  async listMessageSessions(skip: number, top: number, lastUpdatedTime?: Date): Promise<ListSessionsResponse> {
```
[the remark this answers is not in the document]
amarzavery: .net hasn't implemented this yet. they will be doing it soon. However Java has done this.

lib/core/managementClient.ts
```diff
@@ -141,29 +174,61 @@ export class ManagementClient extends LinkEntity {
...
+      messageCount: messageCount
+    });
   }
 
   /**
    * Peeks the desired number of messages from the specified sequence number.
    * @param {Long} fromSequenceNumber The sequence number from where to read the message.
-   * @param {number} [messageCount] The number of messages to retrieve. Default value `1`.
-   * @returns Promise<ReceivedSBMessage[]>
+   * @param {PeekOptions} [options] Options that can be provided while peeking messages.
```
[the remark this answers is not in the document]
amarzavery: The sessionId is an optional parameter for receiveDeferred* method. That is the case with peekBySequenceNumber() as well. However, peekBySequenceNumber has more optional parameters, hence the optional options object, whereas receiveDefered* only has one optional parameter. Creating an options object over there will be an overkill. I don't think there will be any more optional parameters in the near future for receiveDefered* methods

lib/core/managementClient.ts
```diff
@@ -141,29 +174,61 @@ export class ManagementClient extends LinkEntity {
...
+   * Unlike a `received` message, `peeked` message will not have lock token associated with it,
+   * and hence it cannot be `Completed/Abandoned/Deferred/Deadlettered/Renewed`. Also, unlike
+   * `receive() | receiveBatch()` this method will also fetch `Deferred` messages, but
+   * **NOT** `Deadlettered` messages.
+   * @param {string} sessionId The sessionId from which messages need to be peeked.
+   * @param {number} [messageCount] The number of messages to retrieve. Default value `1`.
+   * @returns Promise<ReceivedMessageInfo[]>
+   */
+  async peekMessagesBySession(sessionId: string, messageCount?: number): Promise<ReceivedMessageInfo[]> {
+    if (sessionId == undefined) {
```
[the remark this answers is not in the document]
amarzavery: sessionId is optional in peekBySequenceNumber(), whereas it is required in peekMessagesBySession(). The validation is different.

lib/clientEntityContext.ts
```diff
@@ -29,6 +31,11 @@ export interface ClientEntityContextBase {
    * the connection needs to happen.
    */
   entityPath: string;
+  /**
```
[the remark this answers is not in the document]
ramya-rao-a: Thats good to know, thanks!

lib/clientEntityContext.ts
ramya-rao-a: The SO link is a slightly specific case where the keys are all integers.

In the cases that I have seen so far, especially for _clients in the connection context and the messageSessions here, I would prefer using Map, as it would greatly improve readability of the code.

In most cases, we are iterating over the object and then accessing the element from inside the loop. 

Dictionary:
```
for (const messageSession of Object.keys(this._context.messageSessions)) {
       await this._context.messageSessions[messageSession].close();
}
```



for loop on Map
```
for (var [sessionId, messageSession] of this._context.messageSessions) {
   messageSession.close();
}
```

foreach on Map
```
this._context.messageSessions.forEach(messageSession=> messageSession.close());
```

I am not saying we should make these changes in this PR. Just wanted to get your thoughts on the matter

lib/clientEntityContext.ts
```diff
@@ -89,17 +107,25 @@ export namespace ClientEntityContext {
...
-      if (entityContext.streamingReceiver && entityContext.streamingReceiver.name === name) {
-        result = entityContext.streamingReceiver;
+    (entityContext as ClientEntityContext).sessionManager =
+      new SessionManager(entityContext as ClientEntityContext);
+
+    (entityContext as ClientEntityContext).getReceiver = (name: string, sessionId?: string) => {
+      let receiver: MessageReceiver | MessageSession | undefined = undefined;
+      if (sessionId != undefined && entityContext.messageSessions[sessionId] &&
+        entityContext.messageSessions[sessionId].name === name) {
+        receiver = entityContext.messageSessions[sessionId];
```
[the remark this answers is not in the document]
ramya-rao-a: Well it doesnt really receive the session, but messages from the session. 
Let's keep it as is for now.
I'll take up the naming issue as part of the API reviews

lib/core/managementClient.ts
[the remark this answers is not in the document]
ramya-rao-a: understood

lib/core/managementClient.ts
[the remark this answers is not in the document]
ramya-rao-a: Since neither of these functions are exposed to the user directly, I believe my earlier concern for consistency from the user's perspective is invalid.

lib/serviceBusMessage.ts
ramya-rao-a: I thought so :)

I usually prefer something like `msg.hasOwnProperty('contentType')` in such cases. Thoughts?

lib/queueClient.ts
ramya-rao-a: So user cant control which session they are requesting messages from when using `receiveMessagesFromSessions`?

lib/queueClient.ts
```diff
@@ -309,4 +331,44 @@ export class QueueClient extends Client {
...
   }
+
+  /**
+   * Lists the sessions on the ServiceBus Queue.
+   * @param skip The number of sessions to skip
+   * @param top Maximum numer of sessions.
+   * @param lastUpdateTime Filter to include only sessions updated after a given time. Default
+   * value: 3 days ago from the current time.
+   */
+  async listMessageSessions(skip: number, top: number, lastUpdatedTime?: Date): Promise<ListSessionsResponse> {
```
ramya-rao-a: What about subscription clients? Shouldnt they have session aware features like these? Or do we intend to cover that in a separate PR?

ramya-rao-a: What about subscription clients? Shouldnt they have session aware features like these? Or do we intend to cover that in a separate PR?

lib/session/messageSession.ts
```diff
@@ -0,0 +1,887 @@
...
+
+  /**
+   * Receive a batch of Message objects from a ServiceBus Queue/Topic for a given count and
+   * a given max wait time in seconds, whichever happens first. This method can be used directly
+   * after creating the receiver object and **MUST NOT** be used along with the `start()` method.
+   * @param maxMessageCount The maximum message count. Must be a value greater than 0.
+   * @param maxWaitTimeInSeconds The maximum wait time in seconds for which the Receiver
+   * should wait to receiver the said amount of messages. If not provided, it defaults to 60 seconds.
+   */
+  receiveBatch(maxMessageCount: number, maxWaitTimeInSeconds?: number): Promise<ServiceBusMessage[]> {
```
ramya-rao-a: How about `maxMessageWaitTimeoutInSeconds` just like you added in the `batchingReceiver`?

lib/queueClient.ts
ramya-rao-a: How does one "close" the receiver in this case?

The corresponding `receive` in the QueueClient returns a receiveHandler which is then used to close the receiver.

The only way I can see is to close the queue client itself which seems reasonable, but then begs the question why do we support the "receiveHandler" returned by the `QueueClient.receive`?

lib/session/messageSession.ts
```diff
@@ -0,0 +1,887 @@
...
+        }
+      };
+      // setting the "message" event listener.
+      this._receiver.on(ReceiverEvents.message, onSessionMessage);
+      // adding credit
+      this._receiver!.setCreditWindow(this.maxConcurrentCallsPerSession);
+      this._receiver!.addCredit(this.maxConcurrentCallsPerSession);
+    } else {
+      this._isReceivingMessages = false;
+      const msg = `MessageSession with sessionId '${this.sessionId}' and name '${this.name}' ` +
```
ramya-rao-a: The `recieve` function in both the streamingReceiver.ts and batchReceiver.ts, calls the `_init` function to intialize the receiver when it is not present or it is not open. But here, we throw an error.

lib/session/messageSession.ts
```diff
@@ -0,0 +1,887 @@
...
+      this._notifyError(new Error(msg));
+    }
+  }
+
+  /**
+   * Receive a batch of Message objects from a ServiceBus Queue/Topic for a given count and
+   * a given max wait time in seconds, whichever happens first. This method can be used directly
+   * after creating the receiver object and **MUST NOT** be used along with the `start()` method.
+   * @param maxMessageCount The maximum message count. Must be a value greater than 0.
+   * @param maxWaitTimeInSeconds The maximum wait time in seconds for which the Receiver
```
ramya-rao-a: How about `maxMessageWaitTimeoutInSeconds` just like you added in the `batchingReceiver`?

lib/session/messageSession.ts
```diff
@@ -0,0 +1,887 @@
...
+    (rcvrOptions.source as any).filter[Constants.sessionFilterName] = this.sessionId;
+    return rcvrOptions;
+  }
+
+  /**
+   * Ensures that the session lock is renewed before it expires. The lock will not be renewed for
+   * more than the configured totalAutoLockRenewDuration.
+   * @ignore
+   */
+  private async _ensureSessionLockRenewal(): Promise<void> {
```
ramya-rao-a: Does a session lock renewal also renews the locks the messages in the session?

lib/queueClient.ts
amarzavery: Subscription clients will also have the session aware features. It should be a simple copy/paste of the pattern used in the QueueClient. Hence haven't focused on it in this PR. That can come in a separate PR.

lib/session/messageSession.ts
amarzavery: The behavior for a SessionReceiver is different than a normal MessageReceiver. 
- If the session receiver gets disconnected, then the sdk should not attempt reconnecting the link. It should rather fail and let the user know about it.
- Lock renewal for SessionReceiver happens at the Session level and not for every message in the Session.

Hence the difference in implementation from the MessageReceiver.

lib/session/messageSession.ts
amarzavery: Yup this should be added.

lib/session/messageSession.ts
amarzavery: Yes. As explained in one of the previous comments. Once the link is established then the sdk keeps on renewing the session lock in the background (similar to message lock renewal for message receiver). Renewing the lock for a session takes care of renewing the locks for all the messages in that session.

lib/queueClient.ts
ramya-rao-a: Agreed

lib/session/messageSession.ts
ramya-rao-a: Sweet

lib/queueClient.ts
ramya-rao-a: > Or does this use the same argument as the QueueClient vs MessageReceiver disucssion? i.e .Net meant QueueClient for simple users and MessageSender/MessageReceiver/SessionClient for power users?

Ok, just heard back from Neeraj, that this is indeed the case

lib/queueClient.ts
ramya-rao-a: That makes sense, thanks. 

> If this seems confusing then we can look at creating overloads in TS.... Another option would be to name both the methods differently

The current way of using the options bag works just fine. We just have to ensure that the comments tell the whole story :)



This is exactly why we should have a spec where the the expectations are laid out, instead of every new developer trying to reverse engineer another SDK

lib/session/messageSession.ts
amarzavery: So maxMessageWaitTimeoutInSeconds is an optional property while accepting the messageSession. Thus not needed specifically for receiveBatch

lib/util/constants.ts
amarzavery: sounds good. Shall do that.

lib/core/managementClient.ts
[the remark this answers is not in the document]
ramya-rao-a: Both the older .Net SDK and the current Java SDK have implemented this feature, but without the `skip` and the `top` arguments. Why not stay consistent with them, and add the 2 arguments as per feedback later in the cycle?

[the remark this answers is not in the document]
amarzavery: Was this in the management client (ATOM based XML apis)? Since that is different than the current $management endpoint operation over AMQP.

Was able to find the similar AMQP implementation in java over [here](https://github.com/Azure/azure-service-bus-java/blob/dev/azure-servicebus/src/main/java/com/microsoft/azure/servicebus/primitives/MiscRequestResponseOperationHandler.java#L104).

The only thing that is different is the parameter lastSessionId. This is being set as a property on the request body if present. However, this property is not documented in the [docs](https://github.com/Azure/azure-service-bus-java/blob/dev/azure-servicebus/src/main/java/com/microsoft/azure/servicebus/primitives/MiscRequestResponseOperationHandler.java#L104).

The current implementation as-is works fine, however will have to talk to the service team to get a clear picture on this.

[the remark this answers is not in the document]
amarzavery: Keeping this as-is right now. The current implementation works fine. We can definitely change it later.

amarzavery:
Removing the SessionManager class results in commenting or removing lot of references. It is easier to keep the SessionManager class rather than removing it.

amarzavery:
All comments have been resolved. The CI for previous commit passed. Looks like travis ci is having issues, hence the build is not getting scheduled. The build is passing locally as well. Hence, merging the PR.

PR #17, opened 2018-11-20
 Added maxMessageWaitTimeoutInSeconds optional parameter

amarzavery:
- This ensures that receiveBatch does not wait for a longer time to fetch desired number of messages if no more messages are available.

### DIFF
--- a/examples/simpleSend.ts
+++ b/examples/simpleSend.ts
@@ -4,15 +4,19 @@ dotenv.config();
 
 const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
 const path = process.env.QUEUE_NAME || "";
+const numberOfMessages: number = parseInt((process.argv[2] || "1") as string);
 console.log("str: ", str);
 console.log("path: ", path);
+console.log("Number of messages to send: %d", numberOfMessages);
 
 let ns: Namespace;
 async function main(): Promise<void> {
   ns = Namespace.createFromConnectionString(str);
   const client = ns.createQueueClient(path);
-  await client.send({ body: "Hello sb world!!" + new Date().toString(), messageId: generateUuid() });
-  console.log("***********Created sender and sent the message...");
+  for (let i = 0; i < numberOfMessages; i++) {
+    await client.send({ body: "Hello sb world!!" + new Date().toString(), messageId: generateUuid() });
+    console.log(">>>>>> Sent message number: %d", i + 1);
+  }
 }
 
 main().then(() => {
--- a/lib/core/batchingReceiver.ts
+++ b/lib/core/batchingReceiver.ts
@@ -42,10 +42,17 @@ export class BatchingReceiver extends MessageReceiver {
    *
    * @param {number} maxMessageCount The maximum message count. Must be a value greater than 0.
    * @param {number} [maxWaitTimeInSeconds] The maximum wait time in seconds for which the Receiver
-   * should wait to receiver the said amount of messages. If not provided, it defaults to 60 seconds.
+   * should wait to receive the said amount of messages. If not provided, it defaults to 60 seconds.
+   * @param {number} [maxMessageWaitTimeoutInSeconds] The maximum amount of idle time the Receiver
+   * will wait ater creating the link or after receiving a new message. If no messages are received
+   * in that time frame then the batch receive operation ends. It is advised to keep this value at
+   * 10% of the lockDuration value.
+   * - **Default**: `3` seconds.
    * @returns {Promise<ServiceBusMessage[]>} A promise that resolves with an array of Message objects.
    */
-  receive(maxMessageCount: number, maxWaitTimeInSeconds?: number): Promise<ServiceBusMessage[]> {
+  receive(maxMessageCount: number,
+    maxWaitTimeInSeconds?: number,
+    maxMessageWaitTimeoutInSeconds?: number): Promise<ServiceBusMessage[]> {
     if (!maxMessageCount || (maxMessageCount && typeof maxMessageCount !== 'number')) {
       throw new Error("'maxMessageCount' is a required parameter of type number with a value " +
         "greater than 0.");
@@ -53,6 +60,10 @@ export class BatchingReceiver extends MessageReceiver {
 
     if (maxWaitTimeInSeconds == undefined) {
       maxWaitTimeInSeconds = Constants.defaultOperationTimeoutInSeconds;
+    }
+
+    if (maxMessageWaitTimeoutInSeconds == undefined) {
+      maxMessageWaitTimeoutInSeconds = 3;
     }
 
     const brokeredMessages: ServiceBusMessage[] = [];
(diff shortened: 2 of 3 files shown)


lib/core/batchingReceiver.ts
```diff
@@ -55,6 +62,10 @@ export class BatchingReceiver extends MessageReceiver {
       maxWaitTimeInSeconds = Constants.defaultOperationTimeoutInSeconds;
     }
 
+    if (maxMessageWaitTimeoutInSeconds == undefined) {
+      maxMessageWaitTimeoutInSeconds = 2;
```
amarzavery: I tried 0.1 seconds but it didn't work well. It was too fast and ends up stopping the receiver before starting it. I tried an example with 2 seconds and it worked pretty well. Will stick to this for now. Feel free to change it later based on your findings.

examples/simpleSend.ts
```diff
@@ -4,15 +4,19 @@ dotenv.config();
 
 const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
 const path = process.env.QUEUE_NAME || "";
+const numberOfMessages: number = parseInt((process.argv[2] || "1") as string);
```
AlexGhiondea: Why not keep this coming from the environment as well?

amarzavery: Had that initially. However when testing different scenarios, I found it cumbersome to change the value of the environment variable. Hence came up with this approach of providing the value as a command line arg. 

If consistency is of higher priority then I can switch back to making it an environment variable.

lib/core/batchingReceiver.ts
```diff
@@ -206,6 +240,7 @@ export class BatchingReceiver extends MessageReceiver {
         if (reuse) msg += " Receiver link already present, hence reusing it.";
         log.batching(msg, this._context.namespace.connectionId, maxWaitTimeInSeconds, this.name);
         waitTimer = setTimeout(actionAfterWaitTimeout, (maxWaitTimeInSeconds as number) * 1000);
+        resetTimerOnNewMessageReceived();
```
ramya-rao-a: If I understand this correctlt, `addCreditAndSetTimer` is called at the very beginning, before any messages are recieved.
Why do we need to set the timer here?
Isnt it enough to set the timer after the first message is recieved?

Otherwise, if the first message itself took longer than `maxMessageWaitTimeoutInSeconds`, then we exit and return an empty array to the user.
The user might be wondering why they got an empty array even when the queue has messages and the `maxWaitTimeInSeconds` has not been reached

amarzavery: I see your point. I was thinking about the scenario where there are no active messages in the queue at this moment. Now the link is created and we wait unneccessarily for 60 secs or the specified amount of time.
This ensures that the operation stops asap if it does not receive any messages (first or nth).

What should be the expected behavior?

ramya-rao-a: Arent there any apis to get the number of messages in the queue at any given point in time?

amarzavery: It's the [http management client API](https://github.com/Azure/azure-service-bus-dotnet/blob/1d5df378a5a2d9622ef3118a8bbf3aed549ed34e/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs#L279) that needs to be implemented.

I can add a TODO for now and proceed ahead with waiting after the first message is received.

amarzavery: Another point to think about is: If the first message was delayed due to network latency, then making the management api call would most probably have the same fate. Moreover, it makes things a little complex where 
- what if the first message appeared while the http request was in flight
- what if the making the http request took more time, (node.js doesn't have an out of the box way to cancel an issued http request).

examples/simpleSend.ts
AlexGhiondea: I think it makes sense to keep this like you have it -- as this is a sample it makes sense that you would want to pass around different numbers of messages to put in the queue without using environment variables.

lib/core/batchingReceiver.ts
AlexGhiondea: @amarzavery do you know what the .NET SDK does in that case?

lib/core/messageReceiver.ts
```diff
@@ -315,11 +348,11 @@ export class MessageReceiver extends LinkEntity {
       }
       try {
         await this._onMessage(bMessage);
-        clearTimerAndStopExecution();
+        this._clearMessageLockRenewTimer(bMessage.messageId as string);
```
AlexGhiondea: Should the `clearTimerAndStopExecution` method check to see if a settlement has been started on the message before doing this? Or we assume that once the user's message handler has been called, we can simply remove the renewal lock?

amarzavery: Good question. For the streaming receiver, the customer's message handler is the right place for the customer to settle the message. The message can surely be cached by the customer outside the message handler and the customer may choose to dispose the message at a later stage. However, that would be really odd. 

Have posted this on Teams as well. My understanding of the expected behavior is that either the customer settles the message from within the message handler, or we settle it for her/him (if autoComplete is true).

Currently, the auto message lock renewal timer is cleared whenever
- the action to settle a message has started OR
- the customer's onMessage handler completes execution.

amarzavery:
thats a good point. Infact this also exposes one more condition that needs to be tackled. 
If the customer is taking more time to process the message and the total time exceeds `totalAutoLockRenewDuration` then we should make sure we remove the entry from this. _messageRenewLockTimers Map.

lib/core/messageReceiver.ts
```diff
@@ -307,40 +305,50 @@ export class MessageReceiver extends LinkEntity {
...
+                    bMessage.messageId, err);
+                  // Let the user know that there was an error renewing the message lock.
+                  this._onError!(err);
+                }
 
-            }, amount);
+              }, amount);
+            } else {
+              log.receiver("[%s] Looks like the message lock renew timer has already been " +
+                "cleared for message with id '%s'.", connectionId, bMessage.messageId)
```
ramya-rao-a: Linter is failing due to missing semilcolon here

lib/core/messageReceiver.ts
```diff
@@ -227,6 +244,21 @@ export class MessageReceiver extends LinkEntity {
...
       : 300;
     this.autoRenewLock = this.maxAutoRenewDurationInSeconds > 0 && this.receiveMode === ReceiveMode.peekLock;
+    this._clearMessageLockRenewTimer = (messageId: string) => {
+      if (this._messageRenewLockTimers.get(messageId)) {
+        this._messageRenewLockTimers.get(messageId)!();
+        log.receiver("[%s] Cleared the message renew lock timer for message with id '%s'.",
+          this._context.namespace.connectionId, messageId);
+        this._messageRenewLockTimers.delete(messageId);
+      }
+    };
```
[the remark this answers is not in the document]
amarzavery: That is slick. Thanks for the awesome feedback. Will test the change and update the PR.

lib/core/batchingReceiver.ts
amarzavery: spoke to neeraj and he said that in the .net sdk, maxMessageWaitTimeout gets activated after the receiver link is created. I think it is ok to activate the timer after the first message. We can always tweak it later if we see issues.

PR #18, opened 2018-11-21
Dont wait until defer/abandon is complete to return

ramya-rao-a:


ramya-rao-a:
> Any particular reason to not await?

Because sample for abandon case at the moment throws over 50 error messages. The auto renewal lock code is not aware that the message is abandoned/deferred and so there is it continues to try and fails each time.

I dont think the fix is actually a good one, because as I mentioned before, a user might have long running operations for the duration of which we should be renewing the lock.

I just opened the PR to start the discussion :)

The right fix is for the lock auto-renewal code to be aware of the fact that the message is abandoned/deferred and then skip the renewal altogether.

Perhaps, we can have a property on the message which denotes its state?

amarzavery:
I see. Thanks for providing the full context. Now I see the reason behind this change.
Yes maintaining some kind of state would be the right thing to do. I guess once the action of settling a message was initiated, we should stop auto lock renewal irrespective of wether settling the message succeeds or fails (since a message would only be disposed when processing is complete or cannot be done at this time).

We can have a simple `boolean` property called `settleInitiated` or a property named `settleState` of type `SettleState` which will be initialized in the `ServiceBusMessage` _constructor_
```ts
export interface SettleState {
  settleInitiated: boolean; // default set to false.
  dispositionType: DispositionType;
  succeeded?: boolean; // may be we don't need this. settleInitated && !error => succeeded
  error?: Error | MessagingError;
}
```

Then the condition at [this line](https://github.com/Azure/azure-service-bus-node/blob/master/lib/core/messageReceiver.ts#L284) in the MessageReceiver class can be updated to
```js
if (!bMessage.settleState.settleInitiated && Date.now() < totalAutoLockRenewDuration && continueExecution)
```
IMO, having a proper state that encapsulates every aspect of disposition would be a better approach.
What do you think about it?

Side note: I also discovered a bug in the implementation. If an error happens on the receiver link 
 due to link/session/connection disconnect or any other issue, then while closing the receiver link, we should also clear the autolock renew timer if it exists. I missed doing that.

amarzavery:
The above logic will still not be able to avoid an erroneous state where lock renewal is in flight and the customer initiated settling the message. However this will definitely make sure to not renew the message lock after that. I guess that is acceptable and shouldn't cause a big problem.

AlexGhiondea:
I agree -- I think having state on the message and using that to detect if lock renewal needs to happen seems like the right choice.

I was also wondering if you could also stop the renewal timer once the message starts to be settled -- I would think that once either of abandon/complete/deadletter/defer is called we would, at that time, stop the timer (as well as set the flag). 

That way, we further reduce the window in which the message lock will be renewed after the process of starting the settlement happens.

I think we should also catch the error that comes from the lockRenewal -- if an error happens we can check if the flag to startSettlement is set, and if that is true, then we won't bubble up the error anymore.

Thoughts?

PR #21, opened 2018-11-21
Expose a mechanism for creating a subscription client.

AlexGhiondea:
## Description

Expose a mechanism for creating a subscription client.

/cc @amarzavery @ramya-rao-a


lib/namespace.ts
```diff
@@ -95,6 +95,20 @@ export class Namespace {
     return client;
   }
 
+  /**
+   * Creates a TopicClient for the given topic name. It assumes that the topic has already been
```
ramya-rao-a: SubscriptionClient ?

AlexGhiondea: :)

PR #22, opened 2018-11-21
Add method to get the formatted path for the DeadLetterQueue

### LINKED ISSUE #19: Implement method to return the DeadLetterQueue path for a given queue
Here is the [reference implementation](https://github.com/Azure/azure-service-bus-dotnet/blob/7fabd8e97bcc9bcdea4cc6e0f96ab3f94e030af3/src/Microsoft.Azure.ServiceBus/EntityNameHelper.cs#L28) from the .NET SDK.

AlexGhiondea:
## Description

Add a method on the Namespace to format the DeadLetterPath for a queue.

Fixes #19

### DIFF
--- a/lib/namespace.ts
+++ b/lib/namespace.ts
@@ -195,4 +195,8 @@ export class Namespace {
     const tokenProvider = new AadTokenProvider(credentials);
     return Namespace.createFromTokenProvider(host, tokenProvider, options);
   }
+
+  public static getDeadLetterQueuePathForQueue(queueName: string): string {
+    return `${queueName}/$DeadLetterQueue`;
+  }
 }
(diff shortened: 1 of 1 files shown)


lib/namespace.ts
```diff
@@ -195,4 +195,8 @@ export class Namespace {
     const tokenProvider = new AadTokenProvider(credentials);
     return Namespace.createFromTokenProvider(host, tokenProvider, options);
   }
+
+  public static getDeadLetterQueuePathForQueue(queueName: string): string {
```
amarzavery: in TS all methods on the class are public by default. Hence there is no need to state that. Have been following this convention in all the projects.

AlexGhiondea: Thanks! Fixed.

Issue #12, opened 2018-10-31
Add support for Rules


amarzavery:
- [ ] Add support for implementing rules Docs can be found over [here](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-request-response#rule-operations).
   - [ ] Add Rule
   - [ ] Remove Rule
   - [ ] Get Rules

Take a look at how filters are implemented in [.net sdk](https://github.com/Azure/azure-service-bus-dotnet/tree/dev/src/Microsoft.Azure.ServiceBus/Filters). It will be nice if we can provide TS interfaces for different rules. This will provide a better developer experience.

PR #24, opened 2018-11-29
Implement Topic Filters

ramya-rao-a:
This PR adds support to features listed in #12 i.e Topic Filters

### DIFF
--- a/examples/topics/topicFilters.ts
+++ b/examples/topics/topicFilters.ts
@@ -1,14 +1,15 @@
-import { Namespace, SubscriptionClient } from "../../lib";
+import { Namespace, SubscriptionClient, generateUuid } from "../../lib";
 import * as dotenv from "dotenv";
 dotenv.config();
 
 const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
 const topic = process.env.TOPIC_NAME || "";
-const subscriptionSqlFilter = "with-sql-filter";
-const subscriptionCorrelationFilter = "with-correlation-filter";
+const subscriptionDefaultFilter = process.env.SUBSCRIPTION_NAME_DEFAULT_FILTER || "";
+const subscriptionSqlFilter = process.env.SUBSCRIPTION_NAME_SQL_FILTER || "";
+const subscriptionCorrelationFilter = process.env.SUBSCRIPTION_NAME_CORRELATION_FILTER || "";
 
 let ns: Namespace;
-let client: SubscriptionClient;
+
 const yellowSqlRule = {
   name: "YellowSqlRule",
   expression: "Color = 'Yellow'"
@@ -18,24 +19,63 @@ const correlationRule = {
   filter: { label: "Blue Message" }
 };
 
+const redMessage = {
+  messageId: "RedMessageId" + generateUuid(),
+  body: "Red Message",
+  label: "Red Message",
+  userProperties: {
+    "Color": "Red"
+  }
+};
+
+const blueMessage = {
+  messageId: "BlueMessageId" + generateUuid(),
+  body: "Blue Message",
+  label: "Blue Message",
+  userProperties: {
+    "Color": "Blue"
+  }
+};
+
+const yellowMessage = {
+  messageId: "YelloMessageId" + generateUuid(),
+  body: "Yellow Message",
+  label: "Yellow Message",
+  userProperties: {
+    "Color": "Yellow"
+  }
+};
+
 async function main(): Promise<void> {
   ns = Namespace.createFromConnectionString(str);
 
-  client = ns.createSubscriptionClient(topic, subscriptionSqlFilter);
-  await removeAllRules(client);
-  await client.addSQLRule(yellowSqlRule.name, yellowSqlRule.expression);
-  await testAddedRule(client, yellowSqlRule.name);
-  await client.close();
+  const subscriptionClientNoFilter = ns.createSubscriptionClient(topic, subscriptionDefaultFilter);
+  await removeAllRules(subscriptionClientNoFilter);
+  await subscriptionClientNoFilter.addBooleanRule("$DEFAULT", true);
+  await testAddedRule(subscriptionClientNoFilter, "$DEFAULT");
 
-  client = ns.createSubscriptionClient(topic, subscriptionCorrelationFilter);
-  await removeAllRules(client);
-  await client.addCorrelationRule(correlationRule.name, correlationRule.filter);
-  await testAddedRule(client, correlationRule.name);
-  await client.close();
+  const subscriptionClientSqlFilter = ns.createSubscriptionClient(topic, subscriptionSqlFilter);
+  await removeAllRules(subscriptionClientSqlFilter);
+  await subscriptionClientSqlFilter.addSQLRule(yellowSqlRule.name, yellowSqlRule.expression);
+  await testAddedRule(subscriptionClientSqlFilter, yellowSqlRule.name);
+
+  const subscriptionClientCorrelationFilter = ns.createSubscriptionClient(topic, subscriptionCorrelationFilter);
+  await removeAllRules(subscriptionClientCorrelationFilter);
+  await subscriptionClientCorrelationFilter.addCorrelationRule(correlationRule.name, correlationRule.filter);
+  await testAddedRule(subscriptionClientCorrelationFilter, correlationRule.name);
+
+  await subscriptionClientNoFilter.close();
+  await subscriptionClientSqlFilter.close();
+  await subscriptionClientCorrelationFilter.close();
+
+  const topicClient = ns.createTopicClient(topic);
+  await topicClient.sendBatch([redMessage, blueMessage, yellowMessage]);
+  await topicClient.close();
+
 }
 
 async function removeAllRules(client: SubscriptionClient): Promise<boolean> {
-  const rules = await client.getRules();
+  let rules = await client.getRules();
   console.log(`${rules.length} rules found for ${client.name}`);
 
   for (let i = 0; i < rules.length; i++) {
--- a/examples/topics/topicFilters.ts
+++ b/examples/topics/topicFilters.ts
@@ -1,0 +1,74 @@
+import { Namespace } from "../../lib";
+import * as dotenv from "dotenv";
+import { CorrelationFilter } from '../../lib/core/managementClient';
+dotenv.config();
+
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const topic = process.env.TOPIC_NAME || "";
+const subscription = process.env.SUBSCRIPTION_NAME || "";
+
+let ns: Namespace;
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+  const client = ns.createSubscriptionClient(topic, subscription);
+
+  let rules = await client.getRules();
+  console.log(`${rules.length} rules found for ${client.name}`);
+
+  rules.forEach((rule) => {
+    console.log(`Rule Name: ${rule.name}`);
+    console.log(`Filter: ${JSON.stringify(rule.filter)}`);
+    if (rule.action) {
+      console.log(`Action: ${JSON.stringify(rule.action)}`);
+    }
+  });
+
+  if (rules.find((rule) => rule.name === 'YellowSqlRule')) {
+    await client.removeRule('YellowSqlRule');
+    rules = await client.getRules();
+    if (rules.find((rule) => rule.name === 'YellowSqlRule')) {
+      console.log('Nooooo.... YellowSqlRule should have been removed');
+      return;
+    } else {
+      console.log('Rule YellowSqlRule has been removed');
+    }
+  }
+
+  await client.addSQLRule('YellowSqlRule', "Color = 'Yellow'");
+  rules = await client.getRules();
+  if (rules.find((rule) => rule.name === 'YellowSqlRule')) {
+    console.log('Rule YellowSqlRule has been added');
+  } else {
+    console.log('Nooooo.... Where is the YellowSqlRule rule??');
+    return;
+  }
+
+  if (rules.find((rule) => rule.name === 'ComplicatedRule')) {
+    await client.removeRule('ComplicatedRule');
+    rules = await client.getRules();
+    if (rules.find((rule) => rule.name === 'ComplicatedRule')) {
+      console.log('Nooooo.... ComplicatedRule should have been removed');
+      return;
+    } else {
+      console.log('Rule ComplicatedRule has been removed');
+    }
+  }
+
+  await client.addCorrelationRule('ComplicatedRule', { "correlationId": "some-id", "label": "Blue" });
+  rules = await client.getRules();
+  if (rules.find((rule) => rule.name === 'ComplicatedRule' && (<CorrelationFilter>rule.filter).correlationId === 'some-id' && (<CorrelationFilter>rule.filter).label === 'Blue')) {
+    console.log('Rule ComplicatedRule has been added');
+  } else {
+    console.log('Nooooo.... Where is the ComplicatedRule rule??');
+    return;
+  }
+
+}
+
+main().then(() => {
+  console.log(">>>> Calling close....");
+  return ns.close();
+}).catch((err) => {
+  console.log("error: ", err);
+  return ns.close();
+});
--- a/examples/topics/topicFilters.ts
+++ b/examples/topics/topicFilters.ts
@@ -100,7 +100,7 @@ async function removeAllRules(client: SubscriptionClient): Promise<boolean> {
 }
 
 async function testAddedRule(client: SubscriptionClient, ruleName: string): Promise<boolean> {
-  let rules = await client.getRules();
+  const rules = await client.getRules();
   if (rules.find((rule) => rule.name === ruleName)) {
     console.log(`Rule ${ruleName} has been added`);
   } else {
--- a/examples/topics/topicFilters.ts
+++ b/examples/topics/topicFilters.ts
@@ -1,5 +1,6 @@
 import { Namespace, SubscriptionClient, generateUuid } from "../../lib";
 import * as dotenv from "dotenv";
+import { CorrelationFilter } from '../../lib/core/managementClient';
 dotenv.config();
 
 const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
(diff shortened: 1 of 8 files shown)


lib/core/managementClient.ts
```diff
@@ -532,6 +555,175 @@ export class ManagementClient extends LinkEntity {
...
+
+  /**
+   * Get all the rules on the Subscription.
+   * @returns Promise<RuleDescription> A list of rules.
+   */
+  async getRules(): Promise<RuleDescription[]> {
+    try {
+      const request: AmqpMessage = {
+        body: {
+          top: types.wrap_int(100),
```
ramya-rao-a: @amarzavery Adding a max int value here like the way .Net does, results in the code hanging with the below every 15 seconds. Where is this limitation at? We could ask the user to provide the count, but I was curious to know if you knew anything else about this

image

lib/core/managementClient.ts
```diff
@@ -23,6 +24,28 @@ export enum DispositionStatus {
...
+  "correlation-id"?: string;
+  "message-id"?: string;
+  to?: string;
+  "reply-to"?: string;
+  label?: string;
+  "session-id"?: string;
+  "reply-to-session-id"?: string;
+  "content-type"?: string;
+  properties?: any;
+}
```
ramya-rao-a: In .Net, the above properties use Pascal Casing. I retained the same names as expected by AMQP. This is up for discussion though

lib/core/managementClient.ts
```diff
@@ -532,6 +556,209 @@ export class ManagementClient extends LinkEntity {
...
+      request.application_properties![Constants.trackingId] = generate_uuid();
+
+      log.mgmt("[%s] Get rules request body: %O.", this._context.namespace.connectionId,
+        request.body);
+      log.mgmt("[%s] Acquiring lock to get the management req res link.",
+        this._context.namespace.connectionId);
+      await defaultLock.acquire(this.managementLock, () => { return this._init(); });
+
+      const response = await this._mgmtReqResLink!.sendRequest(request);
+      if (response.application_properties!.statusCode === 204 || !response.body || !Array.isArray(response.body.rules)) {
```
AlexGhiondea: Is this returned when there are no rules defined?

lib/core/managementClient.ts
```diff
@@ -532,6 +556,209 @@ export class ManagementClient extends LinkEntity {
...
+        return [];
+      }
+
+      // Reference: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-request-response#response-11
+      const result: { "rule-description": Typed }[] = response.body.rules || [];
+      const rules: RuleDescription[] = [];
+      result.forEach((x) => {
+        const ruleDescriptor = x['rule-description'];
+        if (!Array.isArray(ruleDescriptor.value) || ruleDescriptor.value.length < 3) {
```
AlexGhiondea: Why 3? Would you mind adding a comment to describe what that means?

lib/core/managementClient.ts
```diff
@@ -532,6 +556,209 @@ export class ManagementClient extends LinkEntity {
...
+          return;
+        }
+        const filtersRawData: Typed = ruleDescriptor.value[0];
+        const actionsRawData: Typed = ruleDescriptor.value[1];
+        const rule: RuleDescription = {
+          name: ruleDescriptor.value[2].value
+        };
+
+        switch (filtersRawData.descriptor.value) {
+          case 83483426823:
```
AlexGhiondea: Could we turn these into constants that are easier to understand?

lib/core/managementClient.ts
```diff
@@ -532,6 +556,209 @@ export class ManagementClient extends LinkEntity {
...
+            if (Array.isArray(filtersRawData.value) && filtersRawData.value.length === 2) {
+              rule.filter = {
+                expression: filtersRawData.value[0] && filtersRawData.value[0].value
+              };
+            }
+            break;
+          case 83483426825:
+            if (Array.isArray(filtersRawData.value) && filtersRawData.value.length === 9) {
+              rule.filter = {
+                correlationId: filtersRawData.value[0] && filtersRawData.value[0].value,
```
AlexGhiondea: Could this be written something like this (to reduce duplication?)

```typescript
function GetValueAtIndex(filtersRawData:Typed, index:number) : any {
    return filtersRawData.value[index] && filtersRawData.value[index].value,
}

// ...

messageId: GetValueAtIndex(filtersRawData, 1);

```

lib/core/managementClient.ts
ramya-rao-a: Yes

lib/core/managementClient.ts
```diff
@@ -532,6 +556,209 @@ export class ManagementClient extends LinkEntity {
...
+                contentType: filtersRawData.value[7] && filtersRawData.value[7].value,
+                userProperties: filtersRawData.value[8] && filtersRawData.value[8].value,
+              };
+            }
+            break;
+          default:
+            break;
+        }
+
+        if (actionsRawData.descriptor.value === 1335734829062
```
AlexGhiondea: I think it would help with code readability if we put these magic numbers in constants.

lib/core/managementClient.ts
ramya-rao-a: The comment with the reference link a few lines above describes the schema of the response object. So I didnt add a specific comment for this.

We use the first 3 elements of the array for filter, action and name of the rule.

They could change the schema in the future to add more details, so I chose to go with < 3 instead of != 3

lib/core/managementClient.ts
```diff
@@ -532,6 +556,209 @@ export class ManagementClient extends LinkEntity {
...
+      };
+      request.application_properties![Constants.trackingId] = generate_uuid();
+
+      log.mgmt("[%s] Remove Rule request body: %O.", this._context.namespace.connectionId,
+        request.body);
+      log.mgmt("[%s] Acquiring lock to get the management req res link.",
+        this._context.namespace.connectionId);
+      await defaultLock.acquire(this.managementLock, () => { return this._init(); });
+
+      await this._mgmtReqResLink!.sendRequest(request);
```
AlexGhiondea: Could this be turned into a boolean-returning function? if the request succeeds we can return true and return false otherwise?

lib/core/managementClient.ts
ramya-rao-a: Yes, but these should live in the amqp-common library. So I have a PR out to that repo. See https://github.com/Azure/amqp-common-js/pull/13

lib/core/managementClient.ts
ramya-rao-a: If it was anything more than a single line of code, then I would agree. In this case, feels like an overkill.

lib/core/managementClient.ts
AlexGhiondea: Got it. I would still add this comment right here in the code so that people don't have to read and understand the docs at the link you specified.

lib/core/managementClient.ts
ramya-rao-a: As soon as https://github.com/Azure/amqp-common-js/pull/13 gets in :)

lib/core/managementClient.ts
AlexGhiondea: Great! Before that PR is accepted, could you use the same code in here? Once the PR is merged and accepted you won't have to touch the code, just consume the new version of the package.

lib/core/managementClient.ts
ramya-rao-a: I had the same thought. Checked with Neeraj. The .NET SDK doesnt return anything either and expects the user to try/catch the call to find the failure case. 

There are other calls in this file which are similar and do not return anything. I decided to keep the same pattern for now and revisit when we have the API Review later

lib/subscriptionClient.ts
```diff
@@ -234,4 +235,60 @@ export class SubscriptionClient extends Client {
...
+    return this._context.managementClient!.removeRule(ruleName);
+  }
+
+  /**
+   * Adds a rule on the subscription as defined by the given rule name, filter and action
+   * @param ruleName Name of the rule
+   * @param filter true or false
+   * @param sqlRuleActionExpression Action to perform if the message satisfies the filtering expression
+   */
+  async addBooleanRule(ruleName: string, filter: boolean, sqlRuleActionExpression?: string): Promise<void> {
```
ramya-rao-a: I am also debating having a single addRule vs multiple. The single one would look like

`async addRule(ruleName, filter: boolean|SQLFilter|CorrelationFilter, sqlRuleActionExpression?: string): Promise<void>`

lib/core/managementClient.ts
ramya-rao-a: We could create new constants for now in this sdk. But we still have to replace them with references to the ones from the amqp library once a new version is released. There is no way we dont have to touch the code again if we want to use the constants from the amqp library

lib/core/managementClient.ts
ramya-rao-a: Done

lib/core/managementClient.ts
ramya-rao-a: Fixed

lib/core/managementClient.ts
ramya-rao-a: Done

lib/core/managementClient.ts
AlexGhiondea: I think we are using that pattern in quite a few places and copy/paste errors could happen -- having a method that we call might be better in the long run.

lib/core/managementClient.ts
AlexGhiondea: I agree -- let's keep this consistent with the other SDKs.

lib/core/managementClient.ts
ramya-rao-a: Done

lib/core/managementClient.ts
```diff
@@ -621,22 +621,22 @@ export class ManagementClient extends LinkEntity {
           case descriptorCodes.sqlFilterList:
             if (Array.isArray(filtersRawData.value) && filtersRawData.value.length === 2) {
               rule.filter = {
-                expression: filtersRawData.value[0] && filtersRawData.value[0].value
+                expression: this._safelyGetTypedValue(filtersRawData.value[0])
```
AlexGhiondea: I wonder if we can use this method in other places in this file (not necessarily) -- see line 651 for a place where we might be able to use this method.

ramya-rao-a: Good catch. Yes, we can use the same there as well.

lib/core/managementClient.ts
```diff
@@ -532,6 +556,224 @@ export class ManagementClient extends LinkEntity {
...
+            break;
+          case descriptorCodes.sqlFilterList:
+            if (Array.isArray(filtersRawData.value) && filtersRawData.value.length === 2) {
+              rule.filter = {
+                expression: this._safelyGetTypedValue(filtersRawData.value[0])
+              };
+            }
+            break;
+          case descriptorCodes.correlationFilterList:
+            if (Array.isArray(filtersRawData.value) && filtersRawData.value.length === 9) {
```
amarzavery: Is it guaranteed to have 9 properties in response (at all the times) even when all the properties are not defined? If not then we will lose the opportunity to deserialize it as a correlationFilterList.

If it does not have 9 properties then we should definitely log an error  (and may be throw an error?)

lib/core/managementClient.ts
```diff
@@ -532,6 +556,224 @@ export class ManagementClient extends LinkEntity {
...
+  }
+
+  /**
+   * Adds a rule on the subscription as defined by the given rule name, filter and action
+   * @param ruleName Name of the rule
+   * @param filter A Boolean, SQL expression or a Correlation filter
+   * @param sqlRuleActionExpression Action to perform if the message satisfies the filtering expression
+   */
+  async addRule(ruleName: string, filter: boolean | string | CorrelationFilter, sqlRuleActionExpression?: string): Promise<void> {
+    if (!ruleName || typeof ruleName !== "string") {
```
amarzavery: Can an empty string be a valid ruleName? If so then we should only check for `typeof ruleName !== "string"`?

Let's assume that an empty string is a valid ruleName. Do we want the sdk to allow that behavior? We should sync up with SB team and code accordingly.

lib/core/managementClient.ts
```diff
@@ -532,6 +556,224 @@ export class ManagementClient extends LinkEntity {
...
+        "endpoint: %O", error);
+      throw error;
+    }
+  }
+
+  /**
+   * Removes the rule on the Subscription identified by the given rule name.
+   * @param ruleName
+   */
+  async removeRule(ruleName: string): Promise<void> {
```
amarzavery:
```js
if (typeof ruleName !== "string") {
  throw new Error("'ruleName' is a required parameter and must be of type 'string'.");
}
```

examples/topics/topicFilters.ts
```diff
@@ -0,0 +1,119 @@
...
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+
+  const subscriptionClientNoFilter = ns.createSubscriptionClient(topic, subscriptionDefaultFilter);
+  await removeAllRules(subscriptionClientNoFilter);
+  await testAddedRule(subscriptionClientNoFilter, subscriptionClientNoFilter.defaultRuleName, true);
+
+  const subscriptionClientSqlFilter = ns.createSubscriptionClient(topic, subscriptionSqlFilter);
+  await removeAllRules(subscriptionClientSqlFilter);
+  await testAddedRule(subscriptionClientSqlFilter, yellowSqlRule.name, yellowSqlRule.expression);
```
amarzavery: Couldn't find an example exercising the sqlRuleActionExpression parameter in the
```js
addRule(ruleName: string, filter: boolean | string | CorrelationFilter, sqlRuleActionExpression?: string)
```

third parameter in the testAddedRule() is an expression. If this is intended to be used as an expression then the example is doing something different. `testAddedRule()` is always adding a rule with ruleName and a filter.

I guess testAddedRule should also have a third optional parameter for sqlRuleExpression and it should have an example for the same.

lib/core/managementClient.ts
```diff
@@ -532,6 +556,224 @@ export class ManagementClient extends LinkEntity {
...
+                label: this._safelyGetTypedValue(filtersRawData.value[4]),
+                sessionId: this._safelyGetTypedValue(filtersRawData.value[5]),
+                replyToSessionId: this._safelyGetTypedValue(filtersRawData.value[6]),
+                contentType: this._safelyGetTypedValue(filtersRawData.value[7]),
+                userProperties: this._safelyGetTypedValue(filtersRawData.value[8]),
+              };
+            }
+            break;
+          default:
+            break;
```
amarzavery: - We should definitely log `filtersRawData.descriptor.value` when we land in the default case with a statement indicating that we have landed in the default case. It can be useful when asking the customer to send debug logs.
- Currently we are dropping the rule description deserialization if it lands in the default case. Should we throw an error instead or may be put the raw value as is?

lib/subscriptionClient.ts
```diff
@@ -234,4 +240,37 @@ export class SubscriptionClient extends Client {
...
+
+
+  /**
+   * Adds a rule on the subscription as defined by the given rule name, filter and action.
+   * Remember to remove the default true filter on the subscription before adding a rule,
+   * otherwise, the added rule will have no affect as the true filter will always result in
+   * the subscription receiving all messages.
+   * @param ruleName Name of the rule
+   * @param filter A Boolean, SQL expression or a Correlation filter
+   * @param sqlRuleActionExpression Action to perform if the message satisfies the filtering expression
```
amarzavery: Since the expression is of type `string`, the customer would have no idea of what would constitute as a valid sql expression. Providing some examples of sqlRuleActionExpression  in a bullet list fashion and a link to the [doc](https://docs.microsoft.com/en-us/azure/service-bus-messaging/topic-filters#actions) would improve the  developer experience.

lib/core/managementClient.ts
```diff
@@ -14,6 +14,7 @@ import { LinkEntity } from "./linkEntity";
 import * as log from "../log";
 import { ReceiveMode } from "./messageReceiver";
 import { reorderLockTokens } from "../util/utils";
+import { TypeDesc, Typed } from 'rhea/typings/types';
```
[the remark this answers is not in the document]
ramya-rao-a: I dont see these types in rhea-promise

lib/core/managementClient.ts
ramya-rao-a: We get a `MessagingError: The specified name ("") is not valid.` error when the rule name is an empty string. Therefore, empty string is not a valid name.

lib/subscriptionClient.ts
ramya-rao-a: Added links to docs

examples/topics/topicFilters.ts
ramya-rao-a: Updated the `testAddedRule()` to include action as a parameter. Our future work on samples will cover the actual sample that will use the action.

lib/core/managementClient.ts
ramya-rao-a: Done

lib/core/managementClient.ts
```diff
@@ -532,6 +556,224 @@ export class ManagementClient extends LinkEntity {
     }
   }
 
+
+  /**
+   * Get all the rules on the Subscription.
+   * @returns Promise<RuleDescription> A list of rules.
```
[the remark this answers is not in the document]
ramya-rao-a: Fixed

lib/core/managementClient.ts
```diff
@@ -14,6 +14,8 @@ import { LinkEntity } from "./linkEntity";
 import * as log from "../log";
 import { ReceiveMode } from "./messageReceiver";
 import { reorderLockTokens } from "../util/utils";
+import { Typed } from 'rhea/typings/types';
+import { max32BitNumber, descriptorCodes } from '../util/constants';
```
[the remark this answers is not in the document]
ramya-rao-a: Fixed

lib/core/managementClient.ts
ramya-rao-a: Logged the case when the descriptor code is not among the expected values.
In such cases, do you use log.mgmt or log.error?

> Currently we are dropping the rule description deserialization if it lands in the default case. Should we throw an error instead or may be put the raw value as is?

The rule description continues, only the filter is dropped.

lib/core/managementClient.ts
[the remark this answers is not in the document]
ramya-rao-a: I dont see these types in rhea-promise...

lib/core/managementClient.ts
```diff
@@ -23,6 +25,28 @@ export enum DispositionStatus {
...
+  filter?: SQLExpression | CorrelationFilter;
+  action?: SQLExpression;
+  name: string;
+}
+
+export interface SQLExpression {
+  expression: string;
+}
+
+export interface CorrelationFilter {
```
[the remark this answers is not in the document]
ramya-rao-a: Fixed

lib/core/managementClient.ts
```diff
@@ -14,6 +14,8 @@ import { LinkEntity } from "./linkEntity";
 import * as log from "../log";
 import { ReceiveMode } from "./messageReceiver";
 import { reorderLockTokens, toBuffer } from "../util/utils";
+import { Typed } from "rhea/typings/types";
```
amarzavery: import { Typed } from "rhea-promise"

lib/core/managementClient.ts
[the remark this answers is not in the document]
amarzavery: Oh I see the real issue now. Typed is an internal type of rhea and is never exposed from rhea itself. Hence, vscode autoimported from the typings folder of rhea. 

So the author of rhea never expected someone to use that type. We need to send a PR to rhea to expose the Typed property and then export it from rhea-promise as well.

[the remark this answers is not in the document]
amarzavery: Another option is to create that Typed interface in our file. Are we exposing the Typed type to our customers?

lib/core/managementClient.ts
```diff
@@ -26,6 +28,79 @@ export enum DispositionStatus {
...
+   */
+  expression: string;
+}
+
+/**
+ * Represents the correlation filter expression.
+ * A CorrelationFilter holds a set of conditions that are matched against one of more of an 
+ * arriving message's user and system properties.
+ */
+export interface CorrelationFilter {
```
amarzavery: We should export all these interfaces from lib/index.ts, so customers can use them for type declaration if required.

lib/core/managementClient.ts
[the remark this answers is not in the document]
ramya-rao-a: No, this is purely for internal use

lib/core/managementClient.ts
ramya-rao-a: Done

lib/core/managementClient.ts
```diff
@@ -23,6 +25,28 @@ export enum DispositionStatus {
...
+  correlationId?: string;
+  messageId?: string;
+  to?: string;
+  replyTo?: string;
+  label?: string;
+  sessionId?: string;
+  replyToSessionId?: string;
+  contentType?: string;
+  userProperties?: any;
```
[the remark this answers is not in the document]
ramya-rao-a: I'll check this when I get to adding tests for the actions and filters

lib/core/managementClient.ts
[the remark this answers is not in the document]
ramya-rao-a: Will use https://github.com/Azure/azure-service-bus-node/issues/29 to track this

PR #26, opened 2018-12-03
Add method for cloning a received message into a message that can be sent to the service

### LINKED ISSUE #20: Implement clone() on a ServiceBusMessage
This would allow re-sending an existing message as a new message.

This is the [reference implementation](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/Message.cs#L299).

AlexGhiondea:
Implement a method that will return a sendable message from a received message.

This will enable re-sending of the same message to the queue.

Fixes #20

### DIFF
--- a/lib/serviceBusMessage.ts
+++ b/lib/serviceBusMessage.ts
@@ -897,4 +897,30 @@ export class ServiceBusMessage implements ReceivedMessage {
       throw new Error(`Cannot find the receiver with name '${this.delivery.link.name}'.`);
     }
   }
+
+  /**
+   * Creates a clone of the current message to allow it to be re-sent to the queue
+   * @returns ServiceBusMessage
+   */
+  clone(): SendableMessageInfo {
+    // We are returning a SendableMessageInfo object because that object can then be sent to ServiceBus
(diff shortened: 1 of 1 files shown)


lib/serviceBusMessage.ts
```diff
@@ -897,4 +897,30 @@ export class ServiceBusMessage implements ReceivedMessage {
       throw new Error(`Cannot find the receiver with name '${this.delivery.link.name}'.`);
     }
   }
+
+  /**
+   * Creates a clone of the current message to allow it to be re-sent to the queue
+   * @returns ServiceBusMessage
+   */
+  clone(): SendableMessageInfo {
+    // We are returning a SendableMessageInfo object because that object can then be sent to ServiceBus
```
ramya-rao-a: I would merge the information in this comment into the jsdocs above

amarzavery: Yup. Putting that information in js docs would be useful.

AlexGhiondea: Done

Issue #31, opened 2018-12-05
Calling receiveBatch twice leads to a hang


AlexGhiondea:
Assuming there are at least 2 messages in the queue, doing the following leads to a hang:

```ts
const msg = await client.receiveBatch(1);
const msg = await client.receiveBatch(1);
```

This is because the batching receiver will create a new batching receiver for every call, without closing the previous one.

amarzavery:
This is as per design. It is expected to not close the batching receiver link. 2nd batch receive call should reuse the link created from the first one if the link is already open. This is done because link creation is expensive operation. Hence we try to minimize link creation wherever possible.

If you received the messages in peeklock mode then you are expected to settle the first message before you can get the second message.

Try setting the receiveMode property to receiveAndDelete and then run the test again OR when you receive the list of first set of messages, please dispose them and then call await receiveBatch again

ramya-rao-a:
> 2nd batch receive call should reuse the link created from the first one if the link is already open.

Agreed. But this isnt what is happening. What is happening is that we are creating a new batchingReceiver for every receiveBatch call.

> If you received the messages in peeklock mode then you are expected to settle the first set of messages before you can get the second set of messages

> Try setting the receiveMode property to receiveAndDelete and then run the test again

Even if we settle the messages using `complete()` before calling the next batch or if we change to receiveAndDelete, it doesnt change the hanging issue which is due to the fact that earlier batchingRecievers never get closed. Only the last one gets closed.

Checkout #32

ramya-rao-a:
Fixed with #32

PR #32, opened 2018-12-05
Allow multiple recieveBatch calls on queueClient

ramya-rao-a:
This PR address the issue #31 

- Updated `receiveBatch` call on the queueClient to re-use existing receiver if it is not currently receiving any messages
- Removed dead code from `finalAction()`.

When working on this, I realized that the initial `onReceiveMessage` handler which is used in the `_init()` call never gets disposed even after calling `this._receiver.removeListener(ReceiverEvents.message, onReceiveMessage);`

Due to this every call to `receiveBatch` will end up calling 2 instances of the `onReceiveMessage`.

@amarzavery Any thoughts on this?

### DIFF
--- a/lib/core/batchingReceiver.ts
+++ b/lib/core/batchingReceiver.ts
@@ -67,7 +67,7 @@ export class BatchingReceiver extends MessageReceiver {
     }
 
     const brokeredMessages: ServiceBusMessage[] = [];
-    let timeOver = false;
+
     this.isReceivingMessages = true;
     return new Promise<ServiceBusMessage[]>((resolve, reject) => {
       let onReceiveMessage: OnAmqpEventAsPromise;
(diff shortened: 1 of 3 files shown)


lib/core/batchingReceiver.ts
```diff
@@ -31,11 +31,11 @@ export class BatchingReceiver extends MessageReceiver {
    * @param {ClientEntityContext} context The client entity context.
    * @param {ReceiveOptions} [options]  Options for how you'd like to connect.
    */
-  constructor(context: ClientEntityContext, options?: ReceiveOptions) {
-    super(context, ReceiverType.batching, options);
-  }
+	constructor(context: ClientEntityContext, options?: ReceiveOptions) {
```
amarzavery: looks like indentation is set to 4 spaces instead of 2

amarzavery: It is difficult to review the actual changes .Can you fix the indentation before the PR can be reviewed?

lib/queueClient.ts
```diff
@@ -13,7 +13,7 @@ import { Client } from "./client";
 import { ReceiveMode, ReceiveOptions, OnError, OnMessage } from "./core/messageReceiver";
 import { ScheduleMessage, ListSessionsResponse } from "./core/managementClient";
 import {
-  MessageSession, AcceptSessionOptions, SessionHandlerOptions, OnSessionMessage
```
amarzavery: Looks like indentation needs to be fixed for all the files in the PR.

lib/core/batchingReceiver.ts
ramya-rao-a: Yes, just did.

PR #34, opened 2018-12-06
Ensure listeners are removed when receivers are re-used

ramya-rao-a:
Since the `onReceiveMessage` of the batchingReciever gets the `.catch()` attached to it before being set as a handler on the receiver, the `removeListener` on it doesnt work which is important when we re-use the subsequent `receiveBatch` calls to work.

This PR covers
- move the try/catch to where the handlers are defined
- avoid passing defaults to `_createReceiverOptions()`

### DIFF
--- a/lib/core/batchingReceiver.ts
+++ b/lib/core/batchingReceiver.ts
@@ -76,6 +76,11 @@ export class BatchingReceiver extends MessageReceiver {
 
     this.isReceivingMessages = true;
     return new Promise<ServiceBusMessage[]>((resolve, reject) => {
+      let onReceiveMessage: OnAmqpEventAsPromise;
+      let onSessionClose: OnAmqpEventAsPromise;
+      let onReceiveClose: OnAmqpEventAsPromise;
+      let onReceiveError: OnAmqpEvent;
+      let onSessionError: OnAmqpEvent;
       let waitTimer: any;
       let maxMessageWaitTimer: any;
       const resetCreditWindow = () => {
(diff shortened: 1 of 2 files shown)


lib/core/batchingReceiver.ts
```diff
@@ -76,15 +76,8 @@ export class BatchingReceiver extends MessageReceiver {
 
     this.isReceivingMessages = true;
     return new Promise<ServiceBusMessage[]>((resolve, reject) => {
-      let onReceiveMessage: OnAmqpEventAsPromise;
-      let onSessionClose: OnAmqpEventAsPromise;
-      let onReceiveClose: OnAmqpEventAsPromise;
-      let onReceiveError: OnAmqpEvent;
-      let onSessionError: OnAmqpEvent;
-      let onSettled: OnAmqpEvent;
```
amarzavery: Due to these tslint errors, declared them here.
```
ERROR: /home/travis/build/Azure/azure-service-bus-node/lib/core/batchingReceiver.ts[92, 71]: variable 'onReceiveError' used before declaration
ERROR: /home/travis/build/Azure/azure-service-bus-node/lib/core/batchingReceiver.ts[93, 65]: variable 'onReceiveMessage' used before declaration
ERROR: /home/travis/build/Azure/azure-service-bus-node/lib/core/batchingReceiver.ts[160, 69]: variable 'onSessionError' used before declaration
```

ramya-rao-a: Yup, just saw that. Fixed

PR #37, opened 2018-12-06
Add sample for bus service for sending/receiving msgs in queue

ShivangiReja:
Port the QueuesGettingStarted sample from the .NET sample.

### DIFF
--- a/samples/queuesGettingStarted.ts
+++ b/samples/queuesGettingStarted.ts
@@ -13,8 +13,7 @@ async function main(): Promise<void> {
 }
 
 async function sendMessage(): Promise<void> {
-  let nsSend: Namespace;
-  nsSend = Namespace.createFromConnectionString(str);
+  const nsSend = Namespace.createFromConnectionString(str);
   const sendClient = nsSend.createQueueClient(path);
   var data = [
     { name: "Einstein", firstName: "Albert" },
@@ -41,20 +40,18 @@ async function sendMessage(): Promise<void> {
     }
     console.log("\n>>>>>> Total Sent messages: %d\n", data.length);
   } catch (err) {
-    console.log("Sending error", err);
+    console.log("Error while sending", err);
   }
   return nsSend.close();
 }
 
 async function receiveMessage(): Promise<void> {
-  let nsRcv: Namespace;
-  nsRcv = Namespace.createFromConnectionString(str);
+  const nsRcv = Namespace.createFromConnectionString(str);
   const receiveClient = nsRcv.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
   try {
     const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
       if (brokeredMessage.label === 'Scientist' &&
         brokeredMessage.contentType === 'application/json') {
-
         console.log("Message Received:", brokeredMessage.body ? brokeredMessage.body.toString() : null);
       }
       await brokeredMessage.complete();
@@ -66,13 +63,13 @@ async function receiveMessage(): Promise<void> {
     await delay(10000);
     await rcvHandler.stop();
   } catch (err) {
-    console.log("Receiving error: ", err);
+    console.log("Error while receiving: ", err);
   }
   return nsRcv.close();
 }
 
 main().then(() => {
-  console.log(">>>> sample Done!!!!");
+  console.log("\n>>>> sample Done!!!!");
 }).catch((err) => {
   console.log("error: ", err);
 });
--- a/samples/queuesGettingStarted.ts
+++ b/samples/queuesGettingStarted.ts
@@ -7,14 +7,13 @@ const path = process.env.QUEUE_NAME || "";
 console.log("str: ", str);
 console.log("path: ", path);
 
-let nsSend: Namespace;
-let nsRcv: Namespace;
 async function main(): Promise<void> {
   await sendMessage();
   await receiveMessage();
 }
 
 async function sendMessage(): Promise<void> {
+  let nsSend: Namespace;
   nsSend = Namespace.createFromConnectionString(str);
   const sendClient = nsSend.createQueueClient(path);
   var data = [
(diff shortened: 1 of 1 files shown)


samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,71 @@
...
+    console.log("Sent message number:", i + 1);
+  }
+  console.log("\n>>>>>> Total Sent messages: %d\n", data.length);
+}
+
+async function receiveMessage(): Promise<void> {
+  nsRcv = Namespace.createFromConnectionString(str);
+  const receiveClient = nsRcv.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    if (brokeredMessage.label != null &&
```
AlexGhiondea: 2 questions:
1. should we compare to null here, or undefined? 
2. should this use `!==`  instead of `!=`?

/cc @ramya-rao-a

samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,71 @@
+import { OnMessage, OnError, MessagingError, delay, ServiceBusMessage, ReceiveMode, generateUuid, Namespace } from "../lib";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const path = process.env.QUEUE_NAME || "";
+console.log("str: ", str);
+console.log("path: ", path);
+
+let nsSend: Namespace;
```
AlexGhiondea: I think these can be moved into the actual methods that use them. Any particular reason to have them as globals here?

samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,71 @@
...
+    var message = {
+      body: JSON.stringify(data[i]),
+      contentType: "application/json",
+      label: "Scientist",
+      messageId: generateUuid(),
+    }
+    await sendClient.send(message);
+    console.log("Sent message number:", i + 1);
+  }
+  console.log("\n>>>>>> Total Sent messages: %d\n", data.length);
```
AlexGhiondea: We should also close the sender when we are done with it.

samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,71 @@
...
+      console.log("Message Received:", brokeredMessage.body ? brokeredMessage.body.toString() : null);
+    }
+    await brokeredMessage.complete();
+  }
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+  const rcvHandler = receiveClient.receive(onMessage, onError, { autoComplete: false });
+  await delay(30000);
+  await rcvHandler.stop();
```
AlexGhiondea: We should close the receiver once we are done with it.

AlexGhiondea:
@ShivangiReja I created a samples folder inside examples -- let's put all the samples there.

samples/queuesGettingStarted.ts
ShivangiReja: Are you talking about the namespaces?

samples/queuesGettingStarted.ts
ShivangiReja: Ok!

samples/queuesGettingStarted.ts
ShivangiReja: Ok!

samples/queuesGettingStarted.ts
AlexGhiondea: Yes -- can those be moved inside the methods that use them?

ShivangiReja: Yes!

samples/queuesGettingStarted.ts
ramya-rao-a: `null != undefined` is false but `null !== undefined` is true. So if `brokeredMessage.label != null` will return false for both when the label is null or undefined.

In this case, we dont need the null check in the first place as the following check on the label text will cover these cases. So both the null checks on the label and the contentType can be removed

samples/queuesGettingStarted.ts
ramya-rao-a: Not if you plan to close the namespace from main().

ramya-rao-a: Also, any reason to use 2 separate instances of the namespace?

ShivangiReja: I'm closing the namespaces inside the methods(send, receive).

ShivangiReja: No specific reason but I was following the .NET code that's why I created 2 different namespaces.

samples/queuesGettingStarted.ts
ShivangiReja: Ok!

samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,71 @@
...
+    { name: "Bohr", firstName: "Niels" },
+    { name: "Faraday", firstName: "Michael" },
+    { name: "Galilei", firstName: "Galileo" },
+    { name: "Kepler", firstName: "Johannes" },
+    { name: "Kopernikus", firstName: "Nikolaus" }
+  ];
+
+  for (let i = 0; i < data.length; i++) {
+    var message = {
+      body: JSON.stringify(data[i]),
```
ramya-rao-a: Since we have set the contentType as json, I wonder if the sending will work without needing the stringification of the json data. Can you test that out?

ShivangiReja: Yes It is working!

samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,71 @@
...
+
+      console.log("Message Received:", brokeredMessage.body ? brokeredMessage.body.toString() : null);
+    }
+    await brokeredMessage.complete();
+  }
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+  const rcvHandler = receiveClient.receive(onMessage, onError, { autoComplete: false });
+  await delay(30000);
```
ramya-rao-a: I would suggest to reduce this delay time, 30 seconds seems like a long time.

samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,71 @@
...
+  await rcvHandler.stop();
+}
+
+main().then(() => {
+  console.log(">>>> Calling close....");
```
ramya-rao-a: The namespace should be closed once we are done here as well as in the catch block

samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,78 @@
...
+
+async function main(): Promise<void> {
+  await sendMessage();
+  await receiveMessage();
+}
+
+async function sendMessage(): Promise<void> {
+  let nsSend: Namespace;
+  nsSend = Namespace.createFromConnectionString(str);
+  const sendClient = nsSend.createQueueClient(path);
```
ramya-rao-a: You can combine the above 2 lines i.e declare and define at the same time.
Also, use `const` instead of `let` as we dont intend to change the value of this variable.

samples/queuesGettingStarted.ts
```diff
@@ -0,0 +1,78 @@
...
+        contentType: "application/json",
+        label: "Scientist",
+        messageId: generateUuid(),
+      }
+      await sendClient.send(message);
+      console.log("Sent message number:", i + 1);
+    }
+    console.log("\n>>>>>> Total Sent messages: %d\n", data.length);
+  } catch (err) {
+    console.log("Sending error", err);
```
ramya-rao-a: For a second I thought we were "sending" the error :)
Change this to "Error while sending"?

samples/queuesGettingStarted.ts
ramya-rao-a: In that case, lets not stringify here. @AlexGhiondea  thoughts?

samples/queuesGettingStarted.ts
ShivangiReja: I referred time delay from the previous samples like simpleSend, receiveComplete etc.

samples/queuesGettingStarted.ts
ShivangiReja: Ok!

samples/queuesGettingStarted.ts
ShivangiReja: Ok!

AlexGhiondea:
@ShivangiReja you need to rebase your changes on top of latest version of master. I can help with that today.

PR #38, opened 2018-12-07
Namespace and Send/Receive tests

ramya-rao-a:
Apart from the tests, this PR covers
- Adding specific error messages for missing names in create queue/topic/subscription client in namespace.ts, because otherwise you get a "entityPath" related error from the clientEntityContext which is too generic. `entityPath` is not the name of the parameter used when creating queue/topic/subscription clients

The sendRecieve tests cover only the "complete" disposition case. Others will be added later on.


test/namespace.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+describe("Create Namespace", function() {
+  describe(".fromConnectionString", function() {
+    it("throws when there is no connection string", function() {
+      testFalsyValues(function(value: any) {
+        const test = function() {
+          return Namespace.createFromConnectionString(value);
+        };
+        test.should.throw(
+          Error,
+          "'connectionString' is a required parameter and must be of type: 'string'."
```
AlexGhiondea: Is this the expected message that it should throw?

test/namespace.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+import { Namespace } from "../lib";
+
+function testFalsyValues(testFn: Function) {
+  [null, undefined, "", 0].forEach(function(value) {
+    testFn(value);
+  });
+}
+
+describe("Create Namespace", function() {
+  describe(".fromConnectionString", function() {
```
AlexGhiondea: Why do we have 2 nested describes?

test/namespace.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+import "mocha";
+import * as chai from "chai";
+const should = chai.should();
+import * as chaiAsPromised from "chai-as-promised";
+import * as dotenv from "dotenv";
+dotenv.config();
+chai.use(chaiAsPromised);
+import { Namespace } from "../lib";
+
+function testFalsyValues(testFn: Function) {
```
AlexGhiondea: Should this be called `testInvalidValues` or something that is more descriptive?

test/namespace.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+  });
+
+  it("throws when sending batch data via a queueClient to a non existing namespace", function() {
+    const client = namespace.createQueueClient("some-name");
+    const sendPromise = client.sendBatch([{ body: "hello" }]).catch(testError);
+
+    return sendPromise.then(() => should.equal(errorWasThrown, true));
+  });
+
+  it("throws when sending batch data via a topicClient to a non existing namespace", function() {
```
AlexGhiondea: Should we split these up based on what they are testing? Queue/Topic/etc?

test/sendReceive.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (peekedSubscriptionMsg.length) {
+      throw new Error("Please use an empty Subscription for integration testing");
+    }
+  });
+
+  afterEach(() => {
+    return namespace.close();
+  });
+
+  it("Simple send and recieveBatch using Queues", async function() {
```
AlexGhiondea: receiveBatch

test/sendReceive.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+  SubscriptionClient
+} from "../lib";
+
+describe("Simple send/receive to/from Queue/Topic/Subscription", function() {
+  let namespace: Namespace;
+  let queueClient: QueueClient;
+  let topicClient: TopicClient;
+  let subscriptionClient: SubscriptionClient;
+
+  beforeEach(async () => {
```
AlexGhiondea: Is there anything special about the queues? As they need to be created prior to the tests running if they need any special setup we should capture it here.

test/namespace.spec.ts
AlexGhiondea: Sounds good!

test/namespace.spec.ts
ramya-rao-a: Fixed

test/sendReceive.spec.ts
ramya-rao-a: Note added

PR #39, opened 2018-12-07
Add SendersReceiversWithQueues and AutoForward sample

AlexGhiondea:
## Description
Add the sample that retrieves from the queue using the receive handler.

### DIFF
--- a/examples/samples/SendersReceiversWithQueues.ts
+++ b/examples/samples/SendersReceiversWithQueues.ts
@@ -63,12 +63,13 @@ async function main(): Promise<void> {
 
   // retrieve all the messages that were sent to the queue (10 messages)
   const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
-    console.log(
-      "### Actual message:",
-      brokeredMessage.body ? brokeredMessage.body.toString() : undefined
-    );
     try {
       if (received <= 10) {
+        console.log(
+          "### Actual message:",
+          brokeredMessage.body ? brokeredMessage.body.toString() : undefined
+        );
+
         // mark the message as completed
         await brokeredMessage.complete();
       }
--- a/examples/samples/SendersReceiversWithQueues.ts
+++ b/examples/samples/SendersReceiversWithQueues.ts
@@ -62,23 +62,20 @@ async function main(): Promise<void> {
 
   // retrieve all the messages that were sent to the queue (10 messages)
   const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
-    try {
+    if (received < 10) {
+      console.log(
+        "### Actual message:",
+        brokeredMessage.body ? brokeredMessage.body.toString() : undefined
+      );
+
+      console.log(`Received: ${received + 1}, message ID: ${brokeredMessage.messageId}`);
+      // we received a message
+      received++;
+
+      // we should only complete the message if we haven't seen 10 messages yet
       if (received < 10) {
-        console.log(
-          "### Actual message:",
-          brokeredMessage.body ? brokeredMessage.body.toString() : undefined
-        );
-
-        // we received a message
-        received++;
-
         // mark the message as completed
         await brokeredMessage.complete();
-      }
-    } finally {
-      if (received === 10) {
-        console.log(`Received ${received} messages.`);
-        rcvHandler.stop();
       }
     }
   };
@@ -88,21 +85,19 @@ async function main(): Promise<void> {
   };
   rcvHandler = client.receive(onMessage, onError, { autoComplete: false });
 
-  // wait 30 seconds, or until we receive 10 messages
-  let timeout = 0;
-  while (received < 10 && timeout < 30 * 1000) {
-    console.log(received);
+  // wait 5 seconds
+  await delay(5000);
 
-    await delay(1000);
-    timeout += 1000;
-  }
+  console.log("Stopping the receiver");
+  await rcvHandler.stop();
 
+  console.log("Closing the client");
   await client.close();
 }
 
 main()
   .then(() => {
-    console.log(">>>> Calling close....");
+    console.log("Closing the namespace");
     return ns.close();
   })
   .catch((err) => {
--- a/examples/samples/SendersReceiversWithQueues.ts
+++ b/examples/samples/SendersReceiversWithQueues.ts
@@ -64,21 +64,21 @@ async function main(): Promise<void> {
   // retrieve all the messages that were sent to the queue (10 messages)
   const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
     try {
-      if (received <= 10) {
+      if (received < 10) {
         console.log(
           "### Actual message:",
           brokeredMessage.body ? brokeredMessage.body.toString() : undefined
         );
 
+        // we received a message
+        received++;
+
         // mark the message as completed
         await brokeredMessage.complete();
       }
     } finally {
-      // we received a message
-      received++;
-
       if (received >= 10) {
-        console.log("Received 10 messages.");
+        console.log(`Received ${received} messages.`);
         rcvHandler.stop();
       }
     }
--- a/examples/samples/SendersReceiversWithQueues.ts
+++ b/examples/samples/SendersReceiversWithQueues.ts
@@ -63,20 +63,17 @@ async function main(): Promise<void> {
   // retrieve all the messages that were sent to the queue (10 messages)
   const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
     if (received < 10) {
-      console.log(
-        "### Actual message:",
-        brokeredMessage.body ? brokeredMessage.body.toString() : undefined
-      );
-
-      console.log(`Received: ${received + 1}, message ID: ${brokeredMessage.messageId}`);
       // we received a message
       received++;
 
-      // we should only complete the message if we haven't seen 10 messages yet
-      if (received < 10) {
-        // mark the message as completed
-        await brokeredMessage.complete();
-      }
+      console.log(
+        `### Received message ${received}: ,${
+          brokeredMessage.body ? brokeredMessage.body.toString() : undefined
+        }`
+      );
+
+      // mark the message as completed
+      await brokeredMessage.complete();
     }
   };
 
--- a/examples/samples/SendersReceiversWithQueues.ts
+++ b/examples/samples/SendersReceiversWithQueues.ts
@@ -1,4 +1,16 @@
-import { ReceiveMode, Namespace, generateUuid, QueueClient, SendableMessageInfo, OnMessage, ServiceBusMessage, OnError, MessagingError, delay, ReceiveHandler } from "../../lib";
+import {
+  ReceiveMode,
+  Namespace,
+  generateUuid,
+  QueueClient,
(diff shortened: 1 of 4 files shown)


examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const path = process.env.QUEUE_NAME || "";
+console.log("str: ", str);
+console.log("path: ", path);
+
+let ns: Namespace;
+
+async function sendMessages(queueClient: QueueClient): Promise<void> {
+  const data: any[] = [
```
ramya-rao-a: This can be `const data: { name: string; firstName: string }[]`

Also, shouldnt `name` here be `lastName`?

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,97 @@
...
+  await sendMessages(client);
+
+  let rcvHandler: ReceiveHandler;
+
+  // Using an object so that we can pass it around by reference
+  const messageCounter = { received: 0 };
+
+  // retrieve all the messages that were sent to the queue (10 messages)
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(">>> Message: ", brokeredMessage);
```
ramya-rao-a: This one is a lot of noise and not very helpful to the user.

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,97 @@
...
+  };
+  rcvHandler = client.receive(onMessage, onError, { autoComplete: false });
+
+  // wait 30 seconds, or until we receive 10 messages
+  await waitForMessagesOrTimeout(messageCounter, 10, 30 * 1000);
+
+  await client.close();
+}
+
+async function waitForMessagesOrTimeout(messageHolder: any, limit: number, timeoutMiliseconds: number): Promise<void> {
```
ramya-rao-a: If we remove this function and inline whats inside it, then we can use a simple number variable for messageCounter instead of the object. Wouldnt that be simpler?

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  // Using an object so that we can pass it around by reference
+  const messageCounter = { received: 0 };
+
+  // retrieve all the messages that were sent to the queue (10 messages)
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(">>> Message: ", brokeredMessage);
+    console.log("### Actual message:", brokeredMessage.body ? brokeredMessage.body.toString() : undefined);
+    try {
+      // mark the message are completed
```
ramya-rao-a: "as" completed

examples/samples/autoForward.ts
```diff
@@ -0,0 +1,73 @@
...
+console.log("topic: ", topic);
+console.log("path: ", path);
+
+let ns: Namespace;
+
+// You need to setup a queue and a topic
+// The topic needs to have a subcription that will forward the messages from the topic to the queue
+
+// The sample assumes the topoloy described here is already setup.
+
```
ramya-rao-a: topoloy ?

examples/samples/autoForward.ts
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  const message2: SendableMessageInfo = {
+    body: "M2",
+    timeToLive: 2 * 60 * 1000, // 2 minutes
+    messageId: generateUuid()
+  };
+
+  console.log("Sending messages");
+  // send 2 duplicated messages
```
ramya-rao-a: But they are not duplicated....

examples/samples/autoForward.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    // complete the message
+    m.complete();
+  }
+
+  if (receivedM1 && receivedM2) {
+    console.log("Successfully received the autoforwarded message");
+  }
+
+  await queueClient.close();
```
ramya-rao-a: Close the topicClient as well?

examples/samples/SendersReceiversWithQueues.ts
AlexGhiondea: Indeed -- interesting how capturing variables along with the await feature makes this work :)

examples/samples/autoForward.ts
AlexGhiondea: Learning to spell a lot of hard words!!! :)

examples/samples/autoForward.ts
AlexGhiondea: Copy/pasta

examples/samples/autoForward.ts
AlexGhiondea: The topicSenderClient is closed at line 40.

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -47,20 +47,19 @@ async function main(): Promise<void> {
   let rcvHandler: ReceiveHandler;
 
   // Using an object so that we can pass it around by reference
-  const messageCounter = { received: 0 };
+  let received = 0;
```
ramya-rao-a: The comment above this line is no longer valid

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  // populate the queue with messages
+  await sendMessages(client);
+
+  let rcvHandler: ReceiveHandler;
+
+  // Using an object so that we can pass it around by reference
+  let received = 0;
+
+  // retrieve all the messages that were sent to the queue (10 messages)
```
ramya-rao-a: This sample doesnt assume that the queue was empty before sending the messages. Meaning, it is very much possible that the queue will have more than 10 messages in a real world scenario when the user is trying out this sample.

In this case, the current code doesnt really ensure that we process only 10 messages.
There are 2 places where we attempt to ensure this.

- Inside the onMessage callback where we call stop() if the count reached 10.
- Outside in the while loop where we wait for 30 seconds or 10 messages whichever comes first

The onMessage callback will be called until the stop() call which is asynchronous completes. Meaning that the 11th, 12th 13th messages will be marked as "complete".
Therefore to ensure that we process only 10 messages, we need to pull the check on the count up top. If count has reached 10, then we call stop() else we complete it.

The outer while loop will exit and close the client, once it sees the count >= 10 and is done sleeping as per the last delay() call. 

At this point, the next `onMessage` callback attempts to "complete" the incoming message and that fails with the below:

image

This is because we attempt to abandon a message if the user's onMessage handler throws an error.

There are 2 action items here:
- Update the `onMessage` callback to not complete any messages after the 10th message
- Dont attempt to abandon the message when user's onMessage handler throws an error for a closed reciever link.

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,110 @@
...
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(
+      "### Actual message:",
+      brokeredMessage.body ? brokeredMessage.body.toString() : undefined
+    );
```
ramya-rao-a: Assuming console.logging the message body is our sample's way of processing the message i.e we can imagine a comment saying "Put your code here", then this should be inside the `if` block which checks the received count.

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,110 @@
...
+    try {
+      if (received <= 10) {
+        // mark the message as completed
+        await brokeredMessage.complete();
```
ramya-rao-a: Say the count is 9.
The `complete()` call sends the request over the wire. Meaning before we hear back the result, the `onMessage` handler can be called for the next message which sees the count to be 9 and we will end up processing/completing this next message.

To solve this, we should put the code that increments the counter before attempting to complete the message

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        // mark the message as completed
+        await brokeredMessage.complete();
+      }
+    } finally {
+      // we received a message
+      received++;
+
+      if (received >= 10) {
+        console.log("Received 10 messages.");
```
ramya-rao-a: Lets use the value of the `received` variable in the console.log here. That would tell the absolute truth :)

ramya-rao-a: Actually on second thoughts, this should be changed to `if (received == 10)`.

We dont want stop() to be called multiple times...

lib/core/messageReceiver.ts
```diff
@@ -436,7 +436,8 @@ export class MessageReceiver extends LinkEntity {
         // Nothing much to do if user's message handler throws. Let us try abandoning the message.
         if (
           error.name !== ConditionErrorNameMapper["com.microsoft:message-lock-lost"] &&
-          this.receiveMode === ReceiveMode.peekLock
+          this.receiveMode === ReceiveMode.peekLock &&
+          this.isOpen() // only try to abandon the messages if the connection is still open
```
ramya-rao-a: @amarzavery FYI

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,110 @@
...
+        // we received a message
+        received++;
+
+        // mark the message as completed
+        await brokeredMessage.complete();
+      }
+    } finally {
+      if (received === 10) {
+        console.log(`Received ${received} messages.`);
+        rcvHandler.stop();
```
ramya-rao-a: Actually.... We still end up inside this if block for the 11th, 12th, 13th... messages because we dont exit early when count is 10

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,105 @@
...
+        "### Actual message:",
+        brokeredMessage.body ? brokeredMessage.body.toString() : undefined
+      );
+
+      console.log(`Received: ${received + 1}, message ID: ${brokeredMessage.messageId}`);
+      // we received a message
+      received++;
+
+      // we should only complete the message if we haven't seen 10 messages yet
+      if (received < 10) {
```
ramya-rao-a: This will result in the last message not being completed

examples/samples/SendersReceiversWithQueues.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (received < 10) {
+      console.log(
+        "### Actual message:",
+        brokeredMessage.body ? brokeredMessage.body.toString() : undefined
+      );
+
+      console.log(`Received: ${received + 1}, message ID: ${brokeredMessage.messageId}`);
+      // we received a message
+      received++;
+
```
ramya-rao-a: To keep things simple, I would re-order the code inside the `if` block to
- Increment the counter
- Then do the console.log. This way we wont be doing the plus 1 in 2 places. Also, skip the messageId in the logs and include the counter in the same statement as that for the message body

AlexGhiondea:
@ramya-rao-a anything else needed for this? Otherwise I'll rebase and merge.

PR #41, opened 2018-12-10
Implement Session Manager

amarzavery:


amarzavery:
The build failure is in node.js version 6.x. May be that version does not support installing a dependency from git. Once the amqp-common package is published, I will update this PR and look at the travis build again.

lib/session/sessionManager.ts
```diff
@@ -62,14 +72,105 @@ export class SessionManager {
...
+          this._maxConcurrentSessionsSemaphore!.release();
+          log.sessionManager("[%s] Releasing the semaphore for max concurrent sessions " +
+            "because an error ocurred.",
+            connectionId
+          );
+        }
+        if (err.name === ConditionErrorNameMapper["amqp:operation-timeout"] ||
+          err.name === ConditionErrorNameMapper["com.microsoft:timeout"] ||
+          err.name === ConditionErrorNameMapper["com.microsoft:session-cannot-be-locked"]) {
+          log.sessionManager("[%s] Sleeping for %d seconds, since there are no more " +
```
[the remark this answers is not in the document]
amarzavery: Yup they will be thrown. Have added a comment in the code based on our discussion.

lib/session/messageSession.ts
```diff
@@ -61,7 +61,6 @@ export interface MessageSessionOptionsBase {
    * @property {number} [maxMessageWaitTimeoutInSeconds] The maximum amount of idle time the session
    * receiver will wait ater a message has been received. If no messages are received in that
    * time frame then the session will be closed.
-   * - **Default**: `60` seconds.
    */
   maxMessageWaitTimeoutInSeconds?: number;
```
amarzavery: Penning the ideas that we agreed upon:
- MessageSession.maxMessageWaitTimeoutInSeconds is set to whatever value provided by the user while accepting a MessageSession. By default its value is undefined.
- It should be explicitly set to `2 seconds` while receiving messages in a batch.
- It should be set explicitly to `60 seconds` while calling MessageSession.receive() from the SessionManager.

lib/session/sessionManager.ts
```diff
@@ -62,14 +72,105 @@ export class SessionManager {
...
+            "active MessageSessions on the ServiceBus entity.",
+            connectionId,
+            noActiveSessionBackOffInSeconds);
+          await delay(noActiveSessionBackOffInSeconds * 1000);
+        } else {
+          onError(err);
+        }
+      } finally {
+        this._maxPendingAcceptSessionsSemaphore!.release();
```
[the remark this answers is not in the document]
amarzavery: resolving based on our discussion. We can have at max let's say 8 concurrent inflight accept session requests. However we can have upto 2000 max concurrent sessions running at any time.

lib/session/messageSession.ts
ramya-rao-a: Currently, in both cases, on timeout, the MessageSession is being closed. But we do not want that to happen when a user is using `receive` or `receiveBatch` after accepting a session.
We only want to close the MessageSession on timeout when using the SessionManager.

lib/session/sessionManager.ts
```diff
@@ -62,13 +72,130 @@ export class SessionManager {
...
+      });
+    }
+  }
+
+  /**
+   * Close the session manager.
+   */
+  close(): void {
+    this._isCancelRequested = true;
+    this.isManagingSessions = false;
```
ramya-rao-a: What is the use case for `isManagingSessions`? It isnt getting used internally. It isnt exposed to the user either.

lib/session/messageSession.ts
ramya-rao-a: Also, why do we want to have this as an option on the MessageSession?
I would expect receiveBatch() to have it as a third parameter just like we do in the batchingReceiver. 
And receive() can have it in its options bag.

lib/session/sessionManager.ts
amarzavery: It is being used in QueueClient.receiveMessgesFromSessions().
The method throws an error if sessionManager is already managingSessions. Thereby restricting the successive calls to receiveMessageFromSessions().

Actually I was thinking to move this check to SessionManager.manageSessions(). Thus we wouldn't need to repeat the same check for subscriptions in SubscriptionClient.

lib/session/messageSession.ts
amarzavery: both the points have been addressed in the latest commit.

lib/session/sessionManager.ts
ramya-rao-a: And we can make it private.

lib/session/sessionManager.ts
```diff
@@ -55,20 +71,170 @@ export class SessionManager {
...
+        log.sessionManager(
+          "[%s] Acquired the semaphore for max pending accept sessions: %d, %d.",
+          connectionId,
+          this._maxPendingAcceptSessionsSemaphore!.currentLockCount(),
+          this._maxPendingAcceptSessionsSemaphore!.awaitedTaskCount()
+        );
+        const messageSession = await MessageSession.create(this._context, {
+          callee: Callee.sessionManager,
+          ...options
+        });
```
ramya-rao-a: Lets close the messageSession here if `this._isCancelRequested` is true.

lib/session/sessionManager.ts
```diff
@@ -55,20 +71,170 @@ export class SessionManager {
...
+            // notify the user about the error.
+            onError(error);
+          }
+        };
+        messageSession.receive(onSessionMessage, onSessionError, options);
       } catch (err) {
-        log.error(err);
+        log.error("[%s] An error occurred while accepting a MessageSession: %O", connectionId, err);
+        if (concurrentSessionSemaphoreAcquired) {
+          concurrentSessionSemaphoreAcquired = false;
```
ramya-rao-a: As discussed, we can remove this variable

PR #77, opened 2018-12-12
Add api-extractor for dts rollups and API reviews

bterlson:
This commit adds API extractor and an npm task `extract-api`. It does
the following:

* Generates a dts rollup file of this library's entire surface area.
  This is presently being saved to /typings/index-rollup.d.ts, but
  should probably go into a different place eventually.
* Adds API baselines under ./etc. Every run of `extract-api` will check
  if the public surface area changed, and if so, say something on the
  console. A git diff of ./etc/service-bus.api.ts shows pretty clearly
  what APIs have changed.

A lot of warnings are generated here. Most seem to be due to different
doc tag conventions, but a few suggest maybe some pieces need to be
exported. Check the generated API file for more details (search for
"WARNING:").

Before accepting this PR, we should at least:

* [ ] Decide if we want to enable the API Review workflow or turn it off
* [ ] Confirm the rollup is accurate and if so replace the typings folder

### DIFF
--- a/etc/service-bus.api.ts
+++ b/etc/service-bus.api.ts
@@ -1,0 +1,411 @@
+// @public
+interface AmqpError {
+  condition?: string;
+  description?: string;
+  info?: any;
+  value?: any[];
+}
+
+// @public
+interface ConnectionConfig {
+}
+
+// @public
+interface CorrelationFilter {
+  contentType?: string;
+  correlationId?: string;
+  label?: string;
+  messageId?: string;
+  replyTo?: string;
+  replyToSessionId?: string;
+  sessionId?: string;
+  to?: string;
+  userProperties?: any;
+}
+
+// @public
+interface DataTransformer {
+  decode: (body: any) => any;
+  encode: (body: any) => any;
+}
+
+// @public
+interface DeadLetterOptions {
+  deadLetterErrorDescription: string;
+  deadletterReason: string;
+}
+
+// @public
+class DefaultDataTransformer implements DataTransformer {
+  decode(body: any): any;
+  encode(body: any): any;
+}
+
+// @public
+export function delay<T>(t: number, value?: T): Promise<T>;
+
+// @public (undocumented)
+interface Delivery {
+  // (undocumented)
+  accept(): void;
+  // (undocumented)
+  readonly format: number;
+  // (undocumented)
+  readonly id: number;
+  // (undocumented)
+  readonly link: Sender | Receiver;
+  // (undocumented)
+  modified(params?: ReleaseParameters): void;
+  // (undocumented)
+  reject(error?: AmqpError): void;
+  // (undocumented)
+  release(params?: ReleaseParameters): void;
+  // (undocumented)
+  readonly remote_settled: boolean;
+  // (undocumented)
+  readonly remote_state?: DeliveryOutcome;
+  // (undocumented)
+  readonly sent: boolean;
+  // (undocumented)
+  readonly settled: boolean;
+  // (undocumented)
+  readonly state?: DeliveryOutcome;
+  // (undocumented)
+  readonly tag: Buffer | string;
+  // (undocumented)
+  update(settled: boolean, state?: any): void;
+}
+
+// @public
+interface Dictionary<T> {
+  // (undocumented)
+  [key: string]: T;
+}
+
+// @public (undocumented)
+interface MessageHandlerOptions {
+  autoComplete?: boolean;
+  maxAutoRenewDurationInSeconds?: number;
+  maxConcurrentCalls?: number;
+}
+
+// @public
+interface MessageHeader {
+  delivery_count?: number;
+  durable?: boolean;
+  first_acquirer?: boolean;
+  priority?: number;
+  ttl?: number;
+}
+
+// @public
+interface MessageProperties {
+  absolute_expiry_time?: number;
+  content_encoding?: string;
+  content_type?: string;
+  correlation_id?: string | number | Buffer;
+  creation_time?: number;
+  group_id?: string;
+  group_sequence?: number;
+  message_id?: string | number | Buffer;
+  reply_to?: string;
+  reply_to_group_id?: string;
+  subject?: string;
+  to?: string;
+  user_id?: string;
+}
+
+// @public
+class MessageSession extends LinkEntity {
+  // WARNING: The type "ClientEntityContext" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "MessageSessionOptions" needs to be exported by the package (e.g. added to index.ts)
+  constructor(context: ClientEntityContext, options?: MessageSessionOptions);
+  autoComplete: boolean;
+  autoRenewLock: boolean;
+  close(): Promise<void>;
+  // WARNING: The type "ClientEntityContext" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "MessageSessionOptions" needs to be exported by the package (e.g. added to index.ts)
+  static create(context: ClientEntityContext, options?: MessageSessionOptions): Promise<MessageSession>;
+  getState(): Promise<void>;
+  isOpen(): boolean;
+  maxAutoRenewDurationInSeconds: number;
+  maxConcurrentCallsPerSession?: number;
+  maxConcurrentSessions?: number;
+  maxMessageWaitTimeoutInSeconds: number;
+  peek(messageCount?: number): Promise<ReceivedMessageInfo[]>;
+  peekBySequenceNumber(fromSequenceNumber: Long, messageCount?: number): Promise<ReceivedMessageInfo[]>;
+  // WARNING: The type "SessionHandlerOptions" needs to be exported by the package (e.g. added to index.ts)
+  receive(onSessionMessage: OnSessionMessage, onError: OnError, options?: SessionHandlerOptions): void;
+  receiveBatch(maxMessageCount: number, maxWaitTimeInSeconds?: number): Promise<ServiceBusMessage[]>;
+  receiveDeferredMessage(sequenceNumber: Long): Promise<ServiceBusMessage | undefined>;
+  receiveDeferredMessages(sequenceNumbers: Long[]): Promise<ServiceBusMessage[]>;
+  receiveMode: ReceiveMode;
+  renewLock(): Promise<Date>;
+  sessionId?: string;
+  sessionLockedUntilUtc?: Date;
+  setState(state: any): Promise<void>;
+  // WARNING: The type "DispositionType" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "DispositionOptions" needs to be exported by the package (e.g. added to index.ts)
+  settleMessage(message: ServiceBusMessage, operation: DispositionType, options?: DispositionOptions): Promise<any>;
+}
+
+// @public
+class MessagingError extends Error {
+  constructor(message: string);
+  condition?: string;
+  info?: any;
+  name: string;
+  retryable: boolean;
+  translated: boolean;
+}
+
+// @public
+class Namespace {
+  constructor(config: ConnectionConfig, options?: NamespaceOptions);
+  close(): Promise<any>;
+  // WARNING: The type "NamespaceOptionsBase" needs to be exported by the package (e.g. added to index.ts)
+  static createFromAadTokenCredentials(host: string, credentials: ApplicationTokenCredentials | UserTokenCredentials | DeviceTokenCredentials | MSITokenCredentials, options?: NamespaceOptions): Namespace;
+  static createFromConnectionString(connectionString: string, options?: NamespaceOptions): Namespace;
+  // WARNING: The type "NamespaceOptionsBase" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "NamespaceOptionsBase" needs to be exported by the package (e.g. added to index.ts)
+  static createFromTokenProvider(host: string, tokenProvider: TokenProvider, options?: NamespaceOptionsBase): Namespace;
+  createQueueClient(queueName: string, options?: QueueClientOptions): QueueClient;
+  createSubscriptionClient(topicName: string, subscriptionName: string, options?: SubscriptionClientOptions): SubscriptionClient;
+  createTopicClient(topicName: string): TopicClient;
+  // (undocumented)
+  static getDeadLetterQueuePathForQueue(queueName: string): string;
+  name: string;
+}
+
+// @public
+interface NamespaceOptions extends NamespaceOptionsBase {
+  tokenProvider?: TokenProvider;
+}
+
+// @public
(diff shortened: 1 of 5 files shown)


etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  subject?: string;
+  to?: string;
+  user_id?: string;
+}
+
+// @public
+class MessageSession extends LinkEntity {
+  // WARNING: The type "ClientEntityContext" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "MessageSessionOptions" needs to be exported by the package (e.g. added to index.ts)
+  constructor(context: ClientEntityContext, options?: MessageSessionOptions);
```
amarzavery: ClientEntityContext need not be exported from the rollup. It is intended for internal use only.
The only way to get a MessageSession instance is to call `acceptSession()` method on the (Queue | Subscription)Client.

amarzavery: MessageSessionOptions need not be exported since `acceptSession(options?: AcceptSessionOptions): Promise<MessageSession>` and `receiveMessgesFromSessions(onSessionMessage: OnSessionMessage, onError: OnError, options?: SessionHandlerOptions): Promise<void>;` on the QueueClient exposes different options that can be provided for creating MessageSession object.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  autoComplete: boolean;
+  autoRenewLock: boolean;
+  close(): Promise<void>;
+  // WARNING: The type "ClientEntityContext" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "MessageSessionOptions" needs to be exported by the package (e.g. added to index.ts)
+  static create(context: ClientEntityContext, options?: MessageSessionOptions): Promise<MessageSession>;
```
amarzavery: same as above.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  getState(): Promise<void>;
+  isOpen(): boolean;
+  maxAutoRenewDurationInSeconds: number;
+  maxConcurrentCallsPerSession?: number;
+  maxConcurrentSessions?: number;
+  maxMessageWaitTimeoutInSeconds: number;
+  peek(messageCount?: number): Promise<ReceivedMessageInfo[]>;
+  peekBySequenceNumber(fromSequenceNumber: Long, messageCount?: number): Promise<ReceivedMessageInfo[]>;
+  // WARNING: The type "SessionHandlerOptions" needs to be exported by the package (e.g. added to index.ts)
```
amarzavery: This has already been exported in a separate PR. Pulling from master branch should get rid of this.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  receiveDeferredMessage(sequenceNumber: Long): Promise<ServiceBusMessage | undefined>;
+  receiveDeferredMessages(sequenceNumbers: Long[]): Promise<ServiceBusMessage[]>;
+  receiveMode: ReceiveMode;
+  renewLock(): Promise<Date>;
+  sessionId?: string;
+  sessionLockedUntilUtc?: Date;
+  setState(state: any): Promise<void>;
+  // WARNING: The type "DispositionType" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "DispositionOptions" needs to be exported by the package (e.g. added to index.ts)
+  settleMessage(message: ServiceBusMessage, operation: DispositionType, options?: DispositionOptions): Promise<any>;
```
amarzavery: settleMessage is intended to be used internally. TypeScript doesn't support declaring an internal access modifier like C#. It only supports public, private, protected. Hence the problem arises. settleMessage() is called from ServiceBusMessage class when the user calls .complete(), .abandon(), .defer() or .deadLetter().

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+
+// @public
+class Namespace {
+  constructor(config: ConnectionConfig, options?: NamespaceOptions);
+  close(): Promise<any>;
+  // WARNING: The type "NamespaceOptionsBase" needs to be exported by the package (e.g. added to index.ts)
+  static createFromAadTokenCredentials(host: string, credentials: ApplicationTokenCredentials | UserTokenCredentials | DeviceTokenCredentials | MSITokenCredentials, options?: NamespaceOptions): Namespace;
+  static createFromConnectionString(connectionString: string, options?: NamespaceOptions): Namespace;
+  // WARNING: The type "NamespaceOptionsBase" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "NamespaceOptionsBase" needs to be exported by the package (e.g. added to index.ts)
```
amarzavery: Yup. This is correct. NamespaceOptionsBase should be exported from index.ts

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  static createFromTokenProvider(host: string, tokenProvider: TokenProvider, options?: NamespaceOptionsBase): Namespace;
+  createQueueClient(queueName: string, options?: QueueClientOptions): QueueClient;
+  createSubscriptionClient(topicName: string, subscriptionName: string, options?: SubscriptionClientOptions): SubscriptionClient;
+  createTopicClient(topicName: string): TopicClient;
+  // (undocumented)
```
amarzavery: This is correct. It should be documented.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  sessionId?: string;
+  timeToLive?: number;
+  to?: string;
+  userProperties?: Dictionary<any>;
+  viaPartitionKey?: string;
+}
+
+// @public
+interface ServiceBusMessageAnnotations extends MessageAnnotations {
+  // WARNING: The name "x-opt-enqueued-time" contains unsupported characters; API names should use only letters, numbers, and underscores
```
amarzavery: This needs to be suppressed. That is how the property is sent on the wire. We cannot change it.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  scheduleMessage(message: SendableMessageInfo, scheduledEnqueueTimeUtc: Date): Promise<Long>;
+  scheduleMessages(messages: ScheduleMessage[]): Promise<Long[]>;
+  send(data: SendableMessageInfo): Promise<Delivery>;
+  sendBatch(datas: SendableMessageInfo[]): Promise<Delivery>;
+}
+
+// WARNING: Unsupported export: generateUuid
+// WARNING: Unsupported export: OnError
+// WARNING: Unsupported export: OnMessage
+// WARNING: Unsupported export: OnSessionMessage
```
amarzavery: I am not sure, what do these comments `Unsupported export` mean?

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+
+// @public
+interface ReceivedMessageInfo {
+}
+
+// @public
+class ReceiveHandler {
+  // WARNING: The type "MessageReceiver" needs to be exported by the package (e.g. added to index.ts)
+  // WARNING: The type "MessageReceiver" needs to be exported by the package (e.g. added to index.ts)
+  constructor(receiver: MessageReceiver);
```
amarzavery: "MessageReceiver" is an internal type and I don't expect this to be used by the customer.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  close(): Promise<any>;
+  // WARNING: The type "ListSessionsResponse" needs to be exported by the package (e.g. added to index.ts)
+  listMessageSessions(skip: number, top: number, lastUpdatedTime?: Date): Promise<ListSessionsResponse>;
+  peek(messageCount?: number): Promise<ReceivedMessageInfo[]>;
+  peekBySequenceNumber(fromSequenceNumber: Long, messageCount?: number): Promise<ReceivedMessageInfo[]>;
+  receive(onMessage: OnMessage, onError: OnError, options?: MessageHandlerOptions): ReceiveHandler;
+  receiveBatch(maxMessageCount: number, maxWaitTimeInSeconds?: number, maxMessageWaitTimeoutInSeconds?: number): Promise<ServiceBusMessage[]>;
+  receiveDeferredMessage(sequenceNumber: Long): Promise<ServiceBusMessage | undefined>;
+  receiveDeferredMessages(sequenceNumbers: Long[]): Promise<ServiceBusMessage[]>;
+  // WARNING: The type "SessionHandlerOptions" needs to be exported by the package (e.g. added to index.ts)
```
amarzavery: This has been fixed and the fix has been checked into master.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+class QueueClient extends Client {
+  // WARNING: The type "ConnectionContext" needs to be exported by the package (e.g. added to index.ts)
+  constructor(name: string, context: ConnectionContext, options?: QueueClientOptions);
+  // WARNING: The type "AcceptSessionOptions" needs to be exported by the package (e.g. added to index.ts)
+  // (undocumented)
+  acceptSession(options?: AcceptSessionOptions): Promise<MessageSession>;
+  cancelScheduledMessage(sequenceNumber: Long): Promise<void>;
+  cancelScheduledMessages(sequenceNumbers: Long[]): Promise<void>;
+  close(): Promise<any>;
+  // WARNING: The type "ListSessionsResponse" needs to be exported by the package (e.g. added to index.ts)
```
amarzavery: "ListSessionsResponse" should be exported.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+}
+
+// @public
+export function parseConnectionString<T>(connectionString: string): ParsedOutput<T>;
+
+// @public (undocumented)
+class QueueClient extends Client {
+  // WARNING: The type "ConnectionContext" needs to be exported by the package (e.g. added to index.ts)
+  constructor(name: string, context: ConnectionContext, options?: QueueClientOptions);
+  // WARNING: The type "AcceptSessionOptions" needs to be exported by the package (e.g. added to index.ts)
```
amarzavery: This has been fixed and the fix has been checked into master.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+  // (undocumented)
```
amarzavery: This has been fixed and the fix has been checked into master.

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+
+// @public
+interface NamespaceOptions extends NamespaceOptionsBase {
+  tokenProvider?: TokenProvider;
+}
+
+// @public
+export function parseConnectionString<T>(connectionString: string): ParsedOutput<T>;
+
+// @public (undocumented)
```
amarzavery:
```js
/**
 * @class QueueClient
 * Defines the operations that can be performed on the Queue after the Queue is created.
 */
```

etc/service-bus.api.ts
```diff
@@ -0,0 +1,411 @@
...
+class DefaultDataTransformer implements DataTransformer {
+  decode(body: any): any;
+  encode(body: any): any;
+}
+
+// @public
+export function delay<T>(t: number, value?: T): Promise<T>;
+
+// @public (undocumented)
+interface Delivery {
```
amarzavery: This type is exported from rhea. Hence the documentation for Delivery in rhea typings should be updated.

amarzavery:
> Decide if we want to enable the API Review workflow or turn it off

How do we turn it off or on ? I see it [here](https://github.com/Azure/azure-service-bus-node/pull/77/files#diff-d207ced746e4bdc3c05d66895c72667fR8) in the api-extractor.json file that apiReviewFile is enabled. Is that what is used to to turn off the api review workflow or is it something else?

> Confirm the rollup is accurate and if so replace the typings folder

One thing that rollup misses is source code documentation comments. Without the source code docs (which can be seen in the typings inside ./typings folder) the typings are less useful.

etc/service-bus.api.ts
AlexGhiondea: Does this mean we need to make a PR into rhea to document this type?

etc/service-bus.api.ts
AlexGhiondea: Does that mean that we need to remove the constructors that take those types? Or make them non-public?

etc/service-bus.api.ts
AlexGhiondea: Do we have a solution for these methods?

etc/service-bus.api.ts
AlexGhiondea: I thought I created the docs for this.. but it looks like I didn't do it correctly -- will take a look at that.

etc/service-bus.api.ts
amarzavery: We can't remove those constructors. I wasnt aware of TS supporting private constructors. Just tried that now. Yes we should definitely make the constructor private.

etc/service-bus.api.ts
amarzavery: Yup.

bterlson:
> How do we turn it off or on ? I see it here in the api-extractor.json file that apiReviewFile is enabled. Is that what is used to to turn off the api review workflow or is it something else?

Yeah, although I suppose you could also just ignore the artifacts it produces, which would allow devs to consult the diff when they needed it.

> One thing that rollup misses is source code documentation comments. Without the source code docs (which can be seen in the typings inside ./typings folder) the typings are less useful.

The docs are in the rollup (not the API diff), I think? Can you point to missing docs? Might be an api-extractor bug.

amarzavery:
> The docs are in the rollup (not the API diff), I think? Can you point to missing docs? Might be an api-extractor bug.

Oh I see. I misread it. Thanks for clearing the confusion. I am also looking at enabling this as a part of autorest.typescript project where the generated sdk will have a dev dependency on api extractor.

As a proof of concept, I tried this locally for ms-rest-js repo. I see lot of errors related to documentation which are noise. Do you know of a way to suppress them?

etc/service-bus.api.ts
ramya-rao-a: `listMessageSessions` has been updated to return an array of string corresponding to the session ids, so this N/A anymore.

etc/service-bus.api.ts
ramya-rao-a: As per recent changes, we no longer have a ReceiveHandler, so N/A.

etc/service-bus.api.ts
ramya-rao-a: `ServiceBusDeliveryAnnotations` and `ServiceBusMessageAnnotations` are not exposed to the user anymore. It wasnt clear how they would get used by the user. So removed until there is a need to have them back

ramya-rao-a:
Closing this PR in favor of https://github.com/Azure/azure-sdk-for-js/pull/1072

We wont be enabling this to run on build yet. Only supporting manual run for now.

Issue #68, opened 2018-12-11
Implement TopicsGettingStarted sample


AlexGhiondea:
https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/TopicsGettingStarted

PR #82, opened 2018-12-17
Add topicGettingStarted sample

ShivangiReja:
Add topicGettingStarted sample #68

### DIFF
--- a/examples/samples/topicsGettingStarted.ts
+++ b/examples/samples/topicsGettingStarted.ts
@@ -21,6 +21,9 @@ const subscription3 = process.env.SUBSCRIPTION_NAME3 || "";
 
 console.log("str: ", str);
 console.log("path: ", topic);
+console.log("Subscription 1: ", subscription1);
+console.log("Subscription 2: ", subscription2);
+console.log("Subscription 3: ", subscription3);
 
 let ns: Namespace;
 async function main(): Promise<void> {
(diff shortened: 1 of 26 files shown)


examples/samples/topicsGettingStarted.ts
```diff
@@ -0,0 +1,96 @@
...
+  const client = ns.createTopicClient(topic);
+
+  const subscription1Client = ns.createSubscriptionClient(topic, subscription1);
+  const subscription2Client = ns.createSubscriptionClient(topic, subscription2);
+  const subscription3Client = ns.createSubscriptionClient(topic, subscription3);
+
+  //retrieving message one by one from each subscription
+  receiveMessage(subscription1Client, "color:red;");
+  receiveMessage(subscription2Client, "color:green;");
+  receiveMessage(subscription3Client, "color:cyan;");
```
ramya-rao-a: What is the purpose of these colors?

ShivangiReja: We are retrieving messages one by one from  each subscription and passing color variable to differentiate the subscription data(just like .Net sample). See the output and Please let me know if I need to change anything.
topicgettingstarted

ramya-rao-a: `client.name`

examples/samples/topicsGettingStarted.ts
```diff
@@ -0,0 +1,95 @@
...
+dotenv.config();
+
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const topic = process.env.TOPIC_NAME || "";
+const subscription1 = process.env.SUBSCRIPTION_NAME1 || "";
+const subscription2 = process.env.SUBSCRIPTION_NAME2 || "";
+const subscription3 = process.env.SUBSCRIPTION_NAME3 || "";
+
+console.log("str: ", str);
+console.log("path: ", topic);
```
AlexGhiondea: I think we should log the 3 subscriptions as well.

ShivangiReja: Ok!

PR #88, opened 2018-12-18
Added README to execute the samples

HarshaNalluru:
## Added README to execute the samples
- Building the library
- Creating a Queue in the Service Bus using Azure Portal
- Executing a sample

### DIFF
--- a/examples/samples/topicsGettingStarted.ts
+++ b/examples/samples/topicsGettingStarted.ts
@@ -1,0 +1,98 @@
+import {
+  OnMessage,
+  OnError,
+  MessagingError,
+  delay,
+  ServiceBusMessage,
+  generateUuid,
+  Namespace,
+  SendableMessageInfo,
+  SubscriptionClient,
+  TopicClient
+} from "../../lib";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const topic = process.env.TOPIC_NAME || "";
+const subscription1 = process.env.SUBSCRIPTION_NAME1 || "";
+const subscription2 = process.env.SUBSCRIPTION_NAME2 || "";
+const subscription3 = process.env.SUBSCRIPTION_NAME3 || "";
+
+console.log("str: ", str);
+console.log("path: ", topic);
+console.log("Subscription 1: ", subscription1);
+console.log("Subscription 2: ", subscription2);
+console.log("Subscription 3: ", subscription3);
+
+let ns: Namespace;
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+  const client = ns.createTopicClient(topic);
+
+  const subscription1Client = ns.createSubscriptionClient(topic, subscription1);
+  const subscription2Client = ns.createSubscriptionClient(topic, subscription2);
+  const subscription3Client = ns.createSubscriptionClient(topic, subscription3);
+
+  //we are setting up receive handlers here
+  setupReceiveHandlers(subscription1Client);
+  setupReceiveHandlers(subscription2Client);
+  setupReceiveHandlers(subscription3Client);
+
+  await sendMessage(client);
+
+  await delay(2000);
+
+  await subscription1Client.close();
+  await subscription2Client.close();
+  await subscription3Client.close();
+}
+
+async function sendMessage(client: TopicClient): Promise<void> {
+  var data = [
+    { name: "Einstein", firstName: "Albert" },
+    { name: "Heisenberg", firstName: "Werner" },
+    { name: "Curie", firstName: "Marie" },
+    { name: "Hawking", firstName: "Steven" },
+    { name: "Newton", firstName: "Isaac" },
+    { name: "Bohr", firstName: "Niels" },
+    { name: "Faraday", firstName: "Michael" },
+    { name: "Galilei", firstName: "Galileo" },
+    { name: "Kepler", firstName: "Johannes" },
+    { name: "Kopernikus", firstName: "Nikolaus" }
+  ];
+  for (let index = 0; index < data.length; index++) {
+    const element = data[index];
+    const message: SendableMessageInfo = {
+      body: `${element.firstName} ${element.name}`,
+      label: "Scientist",
+      timeToLive: 2 * 60 * 1000, // 2 minutes
+      messageId: generateUuid()
+    };
+
+    console.log(`Sending ${message.body} \n`);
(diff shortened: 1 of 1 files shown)


examples/samples/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+- Build the project
+```
+npm run build
+```
+
+## Before executing a sample
+- Go to the Azure Portal(https://portal.azure.com)
+- Here are the docs which would help you create a service bus resource in the portal : https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues
+```
```
ramya-rao-a: For accessibility purposes we should avoid printing the link url directly. Use appropriate link text.

For eg:

`Go to the [Azure Portal](https://portal.azure.com)`

`Follow the docs to [create a Service Bus resource](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues) in the Azure portal.

HarshaNalluru: @ramya-rao-a, Updated the links.

examples/samples/README.md
```diff
@@ -2,42 +2,40 @@
...
-- Here are the docs which would help you create a service bus resource in the portal : https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues
-```
-Go to Dashboard > Service Bus > <your-servicebus-namespace> 
-```
+- Go to the [Azure Portal](https://portal.azure.com)
+- Here are the docs which would help you create a service bus resource in the portal : [ServiceBus - NodeJS DOCS](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues)
+- In the portal, go to **Dashboard > Service Bus > _your-servicebus-namespace_**
 - If you'd like to work with Queues, find the "Queues" tab right under "Entities" at <your-servicebus-namespace> and create a Queue.
-- Go to the Shared access policies under settings tab and add a new one. 
+- Go to the **Shared access policies** under **Settings** tab and add a new one.
```
ramya-rao-a: I dont recall doing this step. I used the default one. Do we really need to do this?

HarshaNalluru: I think you're referring **RootManageSharedAccessKey** as the **default one**. Yes, it is automatically created for the namespace and has permissions for the entire namespace. 

In the [Docs](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas), it is recommended to treat this rule like an administrative root account and don't use it in the application. 

The one I mentioned in the README, iterates the steps to create the Access Keys exclusive to the specific created Queue.

ramya-rao-a: Yes, I was referring to **RootManageSharedAccessKey**.

For the purpose of reducing the number of things a user has to do to run the samples, I would suggest to not have this step as mandatory. This would now need the user to know what a shared access policy is and how to create one.

Here, lets use the **RootManageSharedAccessKey** itself. You can add a side note that this one is like an admin root account and if the user wants to use restricted access they can refer to the doc on using [Shared Access Signatures](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas)

examples/samples/README.md
```diff
@@ -2,42 +2,40 @@
...
 ## Before executing a sample
-- Go to the Azure Portal(https://portal.azure.com)
-- Here are the docs which would help you create a service bus resource in the portal : https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues
-```
-Go to Dashboard > Service Bus > <your-servicebus-namespace> 
-```
+- Go to the [Azure Portal](https://portal.azure.com)
+- Here are the docs which would help you create a service bus resource in the portal : [ServiceBus - NodeJS DOCS](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues)
+- In the portal, go to **Dashboard > Service Bus > _your-servicebus-namespace_**
 - If you'd like to work with Queues, find the "Queues" tab right under "Entities" at <your-servicebus-namespace> and create a Queue.
```
ramya-rao-a: @ShivangiReja Once this PR is checked in, can you add a note here about working with Topic and Subscription, similar to what is there here for Queues?

examples/samples/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## Before executing a sample
+- Go to the [Azure Portal](https://portal.azure.com)
+- Here are the docs which would help you create a service bus resource in the portal : [ServiceBus - NodeJS DOCS](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues)
+- In the portal, go to **Dashboard > Service Bus > _your-servicebus-namespace_**
+- If you'd like to work with Queues, find the "Queues" tab right under "Entities" at <your-servicebus-namespace> and create a Queue.
+- Go to the **Shared access policies** under **Settings** tab and add a new one. 
+- Copy the "Primary Connection String" and the "Name of the Queue" and use them below
+* Example : [Refer this example](https://github.com/Azure/azure-service-bus-node/blob/99b6bff5566f3f8499c8f38a97c7e9f37471cc55/examples/samples/queuesGettingStarted.ts#L14) and update the following lines.
```
ramya-rao-a: The user would have to repeat the copy/paste in all the samples they want to try. An alternative is to create a `.env` file in the root of the project with the below contents
```
SERVICEBUS_CONNECTION_STRING=<Primary-Connection-String>
QUEUE_NAME=<Name-of-the-Queue>
```

examples/samples/README.md
```diff
@@ -0,0 +1,41 @@
...
+```typescript
+const str = process.env.SERVICEBUS_CONNECTION_STRING || <Primary-Connection-String>;
+const path = process.env.QUEUE_NAME || <Name-of-the-Queue>;
+```
+
+## Executing a sample
+- As you've already installed ts-node, you should be able to execute the typescript samples as follows:
```
ramya-rao-a: We earlier mention that `ts-node` is optional. So `As you've already installed` should be `If you have already installed`.

We should then provide an alternative to folks who might not want to use ts-node which would be to use `tsc` to compile the typescript file and the use `node` to run the generated javascript file

examples/samples/README.md
HarshaNalluru: Okay.

examples/samples/README.md
HarshaNalluru: Will update this as well.

Issue #90, opened 2018-12-18
Credit Leak in the batchingReceiver


gcstarr:
**Describe the bug**
When calling receiveBatch sequentially on the same client instance, there is a credit leak if the previous batch received less than maxMessageCount. That is, every time we call receiveBatch on an existing client, it will add maxMessageCount to the credit, even if the previous batch timed-out and used no credit.

**To Reproduce**
- Create a queue Client, and call receiveBatch on a queue with no messages, or at least fewer than maxMessageCount.
- Call receiveBatch again. Observe that the value of maxMessageCount is added as credit even though the previous batch did not use that much.

**Expected behavior**
Only the amount of credit needed to maintain receiving maxMessageCount is added.

**Package-version**: Commit 99b6bff
**node.js version**: 10.14.1
**OS name and version**: Ubuntu 18.10

gcstarr:
As mentioned in #91, the batchingReceiver should be using drain to eliminate unneeded credits (and avoid the above-described issue). I spent some time on this today, and while having some limited success unfortunately ran out of time for now.

ramya-rao-a:
@gcstarr, Please see https://github.com/Azure/azure-service-bus-node/issues/81#issuecomment-460897972

PR #91, opened 2018-12-18
[WIP] Fixing credit leak

gcstarr:
## Description  
When calling receiveBatch sequentially on the same client instance, there is a credit leak if the previous batch received less than maxMessageCount. That is, every time we call receiveBatch on an existing client, it will add maxMessageCount to the credit, even if the previous batch timed-out and used no credit.

- Removing resetCreditWindow as Rhea doesn't do anything for set_credit_window and add_credit calls if the value passed is 0. (see https://github.com/amqp/rhea/blob/290add9fb5806f453874d0cfa71d608b32e3d741/lib/link.js#L371
and
https://github.com/amqp/rhea/blob/290add9fb5806f453874d0cfa71d608b32e3d741/lib/link.js#L359)

- Changing to only add sufficient credit to get credit up to maxMessageCount and no higher, this avoids adding credit above maxMessageCount when the previous batch timed-out and used no credit.
- 

- https://github.com/Azure/azure-service-bus-node/issues/90

### DIFF
--- a/lib/core/batchingReceiver.ts
+++ b/lib/core/batchingReceiver.ts
@@ -301,7 +301,7 @@ export class BatchingReceiver extends MessageReceiver {
         // number of messages concurrently. We will return the user an array of messages that can
         // be of size upto maxMessageCount. Then the user needs to accordingly dispose
         // (complete,/abandon/defer/deadletter) the messages from the array.
-        let creditNeeded: number = Math.max(maxMessageCount - this._receiver!.credit, 0);
+        const creditNeeded: number = Math.max(maxMessageCount - this._receiver!.credit, 0);
         this._receiver!.addCredit(creditNeeded);
         let msg: string = "[%s] Setting the wait timer for %d seconds for receiver '%s'.";
         if (reuse) msg += " Receiver link already present, hence reusing it.";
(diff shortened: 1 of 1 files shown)


amarzavery:
We cannot close the receiver after the messages are returned to the user, since the user may need to settle those messages. We also need to make sure that no more messages are received while the link is active.

Let's say the user asked for 10 messages, however there were only 5 available at that time. Since there were no more messages for the next 2 seconds we we return the list of 5 messages received. Now while the user was processing those messages, 2 more messages were sent to the queue. Since the receiver had 10 credits, it can technically still receive those 2 messages. However, the Promise to receive messages in a batch has already been resolved.

Hence I believe, it is important to let rhea know to stop receiving more messages until further notice. This should always be done at the end of receiveBatch() operation. I think this can be done by draining the receiver of any pending credit so that no new messages can be received.

Check [this issue](https://github.com/amqp/rhea/issues/156) for more explanation.

gcstarr:
Ahh, I think I understand now, thank you. When I get some time (hopefully today) I'll try an approach using drain.

gcstarr:
Closing out this particular pull request as the "drain" approach is the right one.

PR #92, opened 2018-12-18
Added MessageBrowse sample

HarshaNalluru:
## Brief Description
sendMessage() - sends a set of messages
peekMessage() - browse those messages 
receiveMessage() - receive the messages [ **To clear the queue** ]

Main difference from [ .NET Message Browse Sample](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/MessageBrowse) is the addition of  **receiveMessage()** function to clear the flooded queue.

### DIFF
--- a/examples/samples/messageBrowse.ts
+++ b/examples/samples/messageBrowse.ts
@@ -44,7 +44,7 @@ async function sendMessages(queueClient: QueueClient): Promise<void> {
     const message: SendableMessageInfo = {
       body: `${element.firstName} ${element.lastName}`,
       label: "Scientist",
-      timeToLive: 2 * 60 * 1000, // After 2 minutes, the queue gets cleared
+      timeToLive: 2 * 60 * 1000, // After 2 minutes, the message will be removed from the queue
       messageId: generateUuid()
     };
 
@@ -59,7 +59,7 @@ async function peekMessages(): Promise<void> {
   console.log("Browsing messages from Queue...");
 
   try {
-    const count: number = 10;
+    const count = 10;
     const peekedMessage = await receiveClient.peek(count);
     for (let i = 0; i < count; i++) {
       console.log(
--- a/examples/samples/messageBrowse.ts
+++ b/examples/samples/messageBrowse.ts
@@ -1,145 +1,125 @@
-//   
-//   Copyright © Microsoft Corporation, All Rights Reserved
-// 
-//   Licensed under the Apache License, Version 2.0 (the "License"); 
-//   you may not use this file except in compliance with the License. 
-//   You may obtain a copy of the License at
-// 
-//   http://www.apache.org/licenses/LICENSE-2.0 
-// 
-//   THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
-//   OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION
-//   ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A
-//   PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
-// 
-//   See the Apache License, Version 2.0 for the specific language
-//   governing permissions and limitations under the License. 
+import {
+  OnMessage,
+  OnError,
+  MessagingError,
+  delay,
+  ServiceBusMessage,
+  ReceiveMode,
+  generateUuid,
+  Namespace
+} from "../../lib";
+import * as dotenv from "dotenv";
+dotenv.config();
 
-namespace MessageBrowse
-{
-    using Microsoft.Azure.ServiceBus;
-    using Microsoft.Azure.ServiceBus.Core;
-    using Newtonsoft.Json;
-    using System;
-    using System.Text;
-    using System.Threading.Tasks;
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const path = process.env.QUEUE_NAME || "";
 
-    public class Program : MessagingSamples.Sample
-    {
-        public async Task Run(string connectionString)
-        {
-            // send a set of messages
-            await this.SendMessagesAsync(connectionString, BasicQueueName);
-            // broswe those messages 
-            await this.PeekMessagesAsync(connectionString, BasicQueueName);
-        }
+console.log("str: ", str);
+console.log("path: ", path);
 
-        async Task SendMessagesAsync(string connectionString, string queueName)
-        {
-            var sender = new MessageSender(connectionString,queueName);
-            
-            Console.WriteLine("Sending messages to Queue...");
-
-            dynamic data = new[]
-            {
-                new {name = "Einstein", firstName = "Albert"},
-                new {name = "Heisenberg", firstName = "Werner"},
-                new {name = "Curie", firstName = "Marie"},
-                new {name = "Hawking", firstName = "Steven"},
-                new {name = "Newton", firstName = "Isaac"},
-                new {name = "Bohr", firstName = "Niels"},
-                new {name = "Faraday", firstName = "Michael"},
-                new {name = "Galilei", firstName = "Galileo"},
-                new {name = "Kepler", firstName = "Johannes"},
-                new {name = "Kopernikus", firstName = "Nikolaus"}
-            };
-
-            // send a message for each entry in the above array
-            for (int i = 0; i < data.Length; i++)
-            {
-                var message = new Message(Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(data[i])))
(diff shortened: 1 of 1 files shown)


examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,125 @@
...
+console.log("path: ", path);
+
+async function main(): Promise<void> {
+  // send a set of messages
+  await sendMessage();
+  // browse those messages
+  await peekMessage();
+  // receive the messages [ to clear the queue ]
+  await receiveMessage();
+}
```
ramya-rao-a: We are dealing with multiple messages in this sample, but the names of the send/peek/receive suggest that we are dealing with a single message

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,125 @@
...
+    for (let i = 0; i < count; i++) {
+      console.log(
+        `Peeked message ${i}:   
+          ID - ${peekedMessage[i].messageId} , 
+          messageBody - ${peekedMessage[i].body}, 
+          label - ${peekedMessage[i].label}`
+      );
+    }
+    //Uncomment the following statement to check all the attributes of the peekedMessage[0]
+    //console.log(peekedMessage[0]);
```
ramya-rao-a: We want the samples to contain the minimal things that the user needs to get started or minimal things to demonstrate a feature. Therefore, we can remove the above

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,125 @@
...
+
+async function peekMessage(): Promise<void> {
+  const nsRcv = Namespace.createFromConnectionString(str);
+  const receiveClient = nsRcv.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
+  console.log("Browsing messages from Queue...");
+
+  try {
+    const count: number = 10;
+    //peek function takes "the number of messages to retrieve" as the argument
+    //If no arguments are passed to peek(), peeks the first message in the queue [Default value `1`]
```
ramya-rao-a: The documentation of the `peek` method does make this very clear. Any reason on having the comments repeat the docs?

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,125 @@
...
+        body: JSON.stringify(data[i]),
+        contentType: "application/json",
+        label: "Scientist",
+        messageId: generateUuid()
+      };
+      await sendClient.send(message);
+      console.log("Sent message number:", i + 1);
+    }
+    console.log("\n>>>>>>> Total Sent messages: %d\n", data.length);
+  } catch (err) {
```
ramya-rao-a: Since we are mainly worried about the error from the `send` call, we should move the try/catch to just around that.

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,125 @@
...
+    console.log("\n>>>> Browsed the Messages!!!!\n");
+
+    await delay(1000);
+  } catch (err) {
+    console.log("Error while peeking: ", err);
+  }
+  return nsRcv.close();
+}
+
+async function receiveMessage(): Promise<void> {
```
ramya-rao-a: We have other samples demonstrating the receive functionality, do we need to have it in the sample that is meant for just browsing? If the reason is just cleanup, then we can do a simpler `await client.receiveBatch(10)` on a queueClient that has been created in the `ReceiveAndDelete` mode.

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,125 @@
...
+    { name: "Bohr", firstName: "Niels" },
+    { name: "Faraday", firstName: "Michael" },
+    { name: "Galilei", firstName: "Galileo" },
+    { name: "Kepler", firstName: "Johannes" },
+    { name: "Kopernikus", firstName: "Nikolaus" }
+  ];
+  try {
+    for (let i = 0; i < data.length; i++) {
+      var message = {
+        body: JSON.stringify(data[i]),
```
AlexGhiondea: We should use the same method for sending messages like we do in the other samples. I don't think we need to use `JSON.stringify` method here.

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,125 @@
...
+        contentType: "application/json",
```
AlexGhiondea: We might not need to explicitly set the content here.

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,125 @@
...
+        label: "Scientist",
+        messageId: generateUuid()
+      };
+      await sendClient.send(message);
```
AlexGhiondea: This is different from the other samples -- I suggest we use the same pattern here.

examples/samples/messageBrowse.ts
AlexGhiondea: Agreed. I think we should either receive all of them in batch (and complete them) or define a TimeToLive for the message and let the service remove the messages after that time has passed.

examples/samples/messageBrowse.ts
HarshaNalluru: Okay, will remove the two comments.

examples/samples/messageBrowse.ts
HarshaNalluru: Thanks, I'll remove the receiveMessage() function and use the TTL functionality to clear the queue.

examples/samples/messageBrowse.ts
HarshaNalluru: Will change this to the way we send the messages in other samples.

examples/samples/README.md
```diff
@@ -0,0 +1,41 @@
+## Getting started with samples ##
```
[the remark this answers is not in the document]
HarshaNalluru: That's because I created this branch after those commits, I don't think deleting these commits would be a good idea, since that would delete them in the other PR as well. Merging them would be easier because I didn't make any changes to those files in this branch(this PR).

HarshaNalluru:
@AlexGhiondea @ramya-rao-a 
Updated the MessageBrowse sample as per the changes requested in the reviews.
:)

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,86 @@
...
+    { lastName: "Kepler", firstName: "Johannes" },
+    { lastName: "Kopernikus", firstName: "Nikolaus" }
+  ];
+
+  for (let index = 0; index < data.length; index++) {
+    const element = data[index];
+    const message: SendableMessageInfo = {
+      body: `${element.firstName} ${element.lastName}`,
+      label: "Scientist",
+      timeToLive: 2 * 60 * 1000, // After 2 minutes, the queue gets cleared
```
AlexGhiondea: The TimeToLive is on the message, so after 2 minutes the message will be removed from the queue -- if there are messages in the queue with different TTL they won't be cleared. The way the current comment is written it looks like the entire queue will be cleared.

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+
+async function peekMessages(): Promise<void> {
+  const nsRcv = Namespace.createFromConnectionString(str);
+  const receiveClient = nsRcv.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
+  console.log("Browsing messages from Queue...");
+
+  try {
+    const count: number = 10;
```
AlexGhiondea: This can be `const count = 10;`

AlexGhiondea:
@HarshaNalluru I left a couple of very small comments - once those are addressed we can merge this in.

examples/samples/messageBrowse.ts
HarshaNalluru: I see -- will change the comment.

AlexGhiondea:
@HarshaNalluru please rebase on top of latest master and we should be able to merge this in.

examples/samples/messageBrowse.ts
```diff
@@ -0,0 +1,86 @@
...
+    for (let i = 0; i < count; i++) {
+      console.log(
+        `Peeked message ${i}:   
+          ID - ${peekedMessage[i].messageId} , 
+          messageBody - ${peekedMessage[i].body}, 
+          label - ${peekedMessage[i].label}`
+      );
+    }
+    console.log("\n>>>> Browsed the Messages.\n");
+    await delay(1000);
```
ramya-rao-a: What is the reason for this delay?

HarshaNalluru: No need of it. Shall I remove?

ramya-rao-a: Yes please :)

Also rename the `peekedMessage` variable to `peekedMessages` as it points to array of messages and not a single array

PR #107, opened 2018-12-19
Updating batchingReceiver to use drain to empty out credits.

gcstarr:
## Description  
When calling receiveBatch sequentially on the same client instance, there is a credit leak if the previous batch received fewer than maxMessageCount. batchingReceiver should be be setting the drain flow control flag as described in http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-flow-control in order to drain-off unused credits.

Brief description of the changes made in the PR:
- Added onReceiveDrain event handler. The handler unsubscribes the listener and resolves the promise with the received messages (if any).
- Changed finalAction to check remaining credits. If there any, it sets the drain flag on the link to cause the sender to drain the credits. At completion, the onReceiveDrain handler is called. If there are no remaning credits, finalAction directly resolves the promise with the received messages. 

- https://github.com/Azure/azure-service-bus-node/issues/90

### DIFF
--- a/lib/core/batchingReceiver.ts
+++ b/lib/core/batchingReceiver.ts
@@ -101,8 +101,9 @@ export class BatchingReceiver extends MessageReceiver {
 
         if (this._receiver && this._receiver.credit > 0) {
           log.batching(
-            "[%s] Draining leftover credits (%s).",
+            "[%s] Receiver '%s': Draining leftover credits(%d).",
             this._context.namespace.connectionId,
+            this.name,
             this._receiver.credit
           );
 
@@ -116,9 +117,10 @@ export class BatchingReceiver extends MessageReceiver {
 
           this.isReceivingMessages = false;
           log.batching(
-            "[%s] Resolving the promise with received list of messages: %O.",
+            "[%s] Receiver '%s': Resolving receiveBatch() with %d messages.",
             this._context.namespace.connectionId,
-            brokeredMessages
+            this.name,
+            brokeredMessages.length
           );
           resolve(brokeredMessages);
         }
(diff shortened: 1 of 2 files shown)


lib/core/batchingReceiver.ts
```diff
@@ -79,35 +79,40 @@ export class BatchingReceiver extends MessageReceiver {
...
-          this._context.namespace.connectionId,
-          brokeredMessages
-        );
-        resolve(brokeredMessages);
+          if (this._receiver.credit > 0) {
+            log.batching("[%s] Draining leftover credits (%s).", this._context.namespace.connectionId, this._receiver.credit);
+
+            // Setting drain must be accompanied by a flow call (aliased to addCredit in this case).
+            this._receiver.drain = true;
+            this._receiver.addCredit(1);
```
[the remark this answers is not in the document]
gcstarr: It does, but it will be drained off along with the rest of the credit. Rhea will not send a flow frame (addCredit is an alias for flow) if the value passed is 0, so we send 1. Relevant discussion: https://github.com/amqp/rhea/issues/156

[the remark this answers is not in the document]
AlexGhiondea: Thanks for the clarification! 

I would like to have @ramya-rao-a or @amarzavery's review on this as well -- they are currently on vacation. We'll try to get eyes on this early in the new year!

lib/core/batchingReceiver.ts
```diff
@@ -79,35 +79,40 @@ export class BatchingReceiver extends MessageReceiver {
...
+          } else {
+            this._receiver.removeListener(ReceiverEvents.receiverDrained, onReceiveDrain);
+            this.isReceivingMessages = false;
+            log.batching("[%s] Resolving the promise with received list of messages: %O.", this._context.namespace.connectionId, brokeredMessages);
+            resolve(brokeredMessages);
+          }
+        } else {
+          this.isReceivingMessages = false;
```
ramya-rao-a: Why not resolve the messages at this point?

gcstarr: Resolving here would allow the caller to begin a new batch immediately, where the batchingReciever isn't actually ready for a new batch until the receiver_drained event is received.

ramya-rao-a: > Resolving here would allow the caller to begin a new batch immediately, where the batchingReciever isn't actually ready for a new batch until the receiver_drained event is received.

That applies to the `if (this._receiver.credit > 0)` block.

I am referring to the `else` block at line 114. This code path is reached if the receiver is removed from cache i.e `this.receiver` is `null` or `undefined`. At this point there is no way for us to know if there are pending credits or not.

gcstarr: Ah OK, I see. That was a miss on my part, and my manual testing didn't hit that case. I have updated the branch.

PR #108, opened 2018-12-19
Add README to execute the samples 

HarshaNalluru:
Recreating the PR on the right branch
https://github.com/Azure/azure-service-bus-node/pull/88 - Comments from @ramya-rao-a 

Will update the requested changes here.

### DIFF
--- a/examples/samples/README.md
+++ b/examples/samples/README.md
@@ -52,3 +52,9 @@ cd examples\samples
 tsc <sample>.ts
 node <sample>.js
 ```
+- For debugging:
+[VS Code - Debugging](https://code.visualstudio.com/docs/editor/debugging#_launch-configurations) -  In `launch.json`, update the path of envFile and the program to execute the sample.
+```bash
+"program": "${workspaceFolder}/examples/simpleSend.ts"
+"envFile": "${workspaceFolder}/examples/samples/.env"
+```
--- a/examples/samples/README.md
+++ b/examples/samples/README.md
@@ -2,42 +2,40 @@
 
 ## Building the library
 - Clone the repo and cd to the repo directory
-```
+```bash
 git clone https://github.com/azure/azure-service-bus-node.git
 cd azure-service-bus-node
 ```
 - Install typescript, ts-node globally (optional, but very useful)
-```
+```bash
 npm i -g typescript
 npm i -g ts-node
 ```
 - NPM install from the root of the package
-```
+```bash
 npm i
 ```
 - Build the project
-```
+```bash
 npm run build
 ```
 
 ## Before executing a sample
-- Go to the Azure Portal(https://portal.azure.com)
-- Here are the docs which would help you create a service bus resource in the portal : https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues
-```
-Go to Dashboard > Service Bus > <your-servicebus-namespace> 
-```
+- Go to the [Azure Portal](https://portal.azure.com)
+- Here are the docs which would help you create a service bus resource in the portal : [ServiceBus - NodeJS DOCS](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues)
+- In the portal, go to **Dashboard > Service Bus > _your-servicebus-namespace_**
 - If you'd like to work with Queues, find the "Queues" tab right under "Entities" at <your-servicebus-namespace> and create a Queue.
-- Go to the Shared access policies under settings tab and add a new one. 
+- Go to the **Shared access policies** under **Settings** tab and add a new one. 
 - Copy the "Primary Connection String" and the "Name of the Queue" and use them below
-```
-Example : (https://github.com/Azure/azure-service-bus-node/blob/99b6bff5566f3f8499c8f38a97c7e9f37471cc55/examples/samples/queuesGettingStarted.ts#L14).
+* Example : [Refer this example](https://github.com/Azure/azure-service-bus-node/blob/99b6bff5566f3f8499c8f38a97c7e9f37471cc55/examples/samples/queuesGettingStarted.ts#L14) and update the following lines.
+```typescript
 const str = process.env.SERVICEBUS_CONNECTION_STRING || <Primary-Connection-String>;
 const path = process.env.QUEUE_NAME || <Name-of-the-Queue>;
 ```
 
 ## Executing a sample
 - As you've already installed ts-node, you should be able to execute the typescript samples as follows:
-```
+```bash
 cd examples\samples
 ts-node <sample>.ts
 ```
--- a/examples/samples/README.md
+++ b/examples/samples/README.md
@@ -24,19 +24,35 @@ npm run build
 - Go to the [Azure Portal](https://portal.azure.com).
 - Here are the docs which would help you create a service bus resource in the portal : [ServiceBus - NodeJS DOCS](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues).
 - In the portal, go to **Dashboard > Service Bus > _your-servicebus-namespace_**.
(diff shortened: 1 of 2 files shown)


AlexGhiondea:
@HarshaNalluru are there comments in the previous PR that still need to be addressed?

HarshaNalluru:
Yes, yet to make some changes. will inform when I'm done.

AlexGhiondea:
@HarshaNalluru how is this going?

HarshaNalluru:
As Ramya advised, working on the env variables.. will push after that's done.

HarshaNalluru:
> _Example: [**dotenv.config();**](https://github.com/Azure/azure-service-bus-node/blob/99b6bff5566f3f8499c8f38a97c7e9f37471cc55/examples/samples/queuesGettingStarted.ts#L12) call in the sample loads the strings in `.env` into `process.env`._

HarshaNalluru:
Done @AlexGhiondea!

HarshaNalluru:
@ShivangiReja 
Check this and let me know if you've things to add.

ShivangiReja:
@HarshaNalluru  There is one more option to execute the sample- Change launch.json file
* If you want to run the current file add this: `"program": "${file}",`
* otherwise add full path of the file : "program": "${workspaceFolder}/examples/simpleSend.ts",

examples/samples/README.md
```diff
@@ -0,0 +1,60 @@
...
+npm run build
+```
+
+## Before executing a sample
+- Go to the [Azure Portal](https://portal.azure.com).
+- Here are the docs which would help you create a service bus resource in the portal : [ServiceBus - NodeJS DOCS](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues).
+- In the portal, go to **Dashboard > Service Bus > _your-servicebus-namespace_**.
+- Copy the "Primary Connection String" of **RootManageSharedAccessKey** at **Shared access policies** under **Settings** tab.
+- To work with Queues, find the "Queues" tab right under "Entities" at **_your-servicebus-namespace_** and create a Queue and copy its name for next steps.
+- To work with Topics, find the "Topics" tab right under "Entities" at **_your-servicebus-namespace_**, create a Topic. Go to **_your-servicebus-namespace_ > _your-topic_**, create subscriptions for the topic and note down their names for the next step.
```
ramya-rao-a: The 3 points above are all about noting down values from the portal for later use, but they are not consistent. 2 of them use the term "copy", the other uses the term "note down". Let's keep them consistent and use the term "note down" to keep things clear

examples/samples/README.md
```diff
@@ -0,0 +1,60 @@
...
+```bash
+npm run build
+```
+
+## Before executing a sample
+- Go to the [Azure Portal](https://portal.azure.com).
+- Here are the docs which would help you create a service bus resource in the portal : [ServiceBus - NodeJS DOCS](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues).
+- In the portal, go to **Dashboard > Service Bus > _your-servicebus-namespace_**.
+- Copy the "Primary Connection String" of **RootManageSharedAccessKey** at **Shared access policies** under **Settings** tab.
+- To work with Queues, find the "Queues" tab right under "Entities" at **_your-servicebus-namespace_** and create a Queue and copy its name for next steps.
```
ramya-rao-a: Break this sentence up for ease of reading.

`... and create a Queue. Copy its name for next steps.` instead of `... and create a Queue and copy its name for next steps`

examples/samples/README.md
```diff
@@ -0,0 +1,60 @@
...
+- To work with Topics, find the "Topics" tab right under "Entities" at **_your-servicebus-namespace_**, create a Topic. Go to **_your-servicebus-namespace_ > _your-topic_**, create subscriptions for the topic and note down their names for the next step.
+> _Note : **RootManageSharedAccessKey** is automatically created for the namespace and has permissions for the entire namespace. If you want to use restricted access, refer [Shared Access Signatures](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas), create the Access Keys exclusive to the specific created Queue/Topic._
+
```
ramya-rao-a: Perfect :)

examples/samples/README.md
```diff
@@ -0,0 +1,60 @@
...
+Create a `.env` file in the `examples/samples` folder with the following contents. Replace the `<string>` with your corresponding values.
+```typescript
+SERVICEBUS_CONNECTION_STRING=<Primary-Connection-String>
+QUEUE_NAME=<Name-of-the-Queue>
+TOPIC_NAME=<Name-of-the-Topic>
+SUBSCRIPTION_NAME1=<Subscription-1>
+SUBSCRIPTION_NAME2=<Subscription-2>
+SUBSCRIPTION_NAME3=<Subscription-3>
+```
+> _Example: [**dotenv.config();**](https://github.com/Azure/azure-service-bus-node/blob/99b6bff5566f3f8499c8f38a97c7e9f37471cc55/examples/samples/queuesGettingStarted.ts#L12) call in the sample loads the strings in `.env` into `process.env`._
```
ramya-rao-a: We can skip the "example" part of this statement and just mention that these values will get loaded into `process.env`.

We can also combine this line to before the .env file contents

examples/samples/README.md
```diff
@@ -0,0 +1,60 @@
...
+ts-node <sample>.ts
+```
+- Otherwise, execute them in the following manner:
+```bash
+cd examples\samples
+tsc <sample>.ts
+node <sample>.js
+```
+- For debugging:
+[VS Code - Debugging](https://code.visualstudio.com/docs/editor/debugging#_launch-configurations) -  In `launch.json`, update the path of envFile and the program to execute the sample.
```
ramya-rao-a: launch.json will have multiple debug configurations. Mention which config should be updated.

examples/samples/README.md
```diff
@@ -0,0 +1,60 @@
...
+```bash
+"program": "${workspaceFolder}/examples/simpleSend.ts"
+"envFile": "${workspaceFolder}/examples/samples/.env"
```
ramya-rao-a: Can we not add this `envFile` to the launch.json file right way and not require the user to do so?

HarshaNalluru: There is one already with this path `"${workspaceFolder}/.env"` in "launch.json". 
Samples in the `${workspaceFolder}/examples`(not  `"${workspaceFolder}/examples/samples"`) are using the path.
If we don't change it to `"${workspaceFolder}/examples/samples/.env"`, we would have create .env outside the `examples/samples` and update `dotenv.config();` in the samples inside  `examples/samples` by passing the path of `.env` file, i.e., `dotenv.config({ path: "../../.env" });`.

ramya-rao-a: > If we don't change it to "${workspaceFolder}/examples/samples/.env", we would have create .env outside the examples/samples and update dotenv.config(); in the samples(inside examples/samples) by passing the path of .env file, i.e., dotenv.config({ path: "../../.env" });.

`dotenv.config()` picks up the .env file at the root of the project. Why dont we just use that instead of creating a new one under the samples folder?

HarshaNalluru: Even I would recommend .env file at the root of the project. Then we'd have to call in this manner `dotenv.config({ path: "../../.env" });`.  

> `dotenv.config()` doesn't pick up .env at the root, the implementation of it seems that it looks for .env in the current directory.

ramya-rao-a: I am able to debug the samples using the existing debug configuration and updating just the `program` attribute.

I am also able to debug the same after removing the `envFile` attribute. This is because `dotenv` is doing the right thing by reading the .env file under $workspaceRoot

HarshaNalluru:
@ramya-rao-a 
This is not done yet. After the discussion, `.env` is supposed to be created at the root folder. We need to update that and accordingly, at the time of execution, we need to execute the samples(using `ts-node`) from the root folder.

ramya-rao-a:
Good catch, please create a new PR for these changes

PR #111, opened 2018-12-20
Deadletter fix

### LINKED ISSUE #110: Fix deadLetter() issue.
When deadlettering a message without specifying the optional parameter in the method call you are getting error-- The required field 'condition' is not set in 'amqp:error:list'.

ShivangiReja:
Fixing issue when user has not passed DeadLetterOptions in deadletter method.
Fixes #110

### DIFF
--- a/lib/serviceBusMessage.ts
+++ b/lib/serviceBusMessage.ts
@@ -940,12 +940,10 @@ export class ServiceBusMessage implements ReceivedMessage {
     options: DeadLetterOptions = { deadletterReason: "", deadLetterErrorDescription: "" }
   ): Promise<void> {
     let error: AmqpError = {};
-    if (options) {
-      error = {
-        condition: options.deadletterReason,
-        description: options.deadLetterErrorDescription
-      };
-    }
+    error = {
+      condition: options.deadletterReason,
+      description: options.deadLetterErrorDescription
+    };
     log.message(
       "[%s] Deadlettering the message with id '%s'.",
       this._context.namespace.connectionId,
--- a/lib/serviceBusMessage.ts
+++ b/lib/serviceBusMessage.ts
@@ -943,6 +943,11 @@ export class ServiceBusMessage implements ReceivedMessage {
         condition: options.deadletterReason,
         description: options.deadLetterErrorDescription
       };
+    } else {
+      error = {
+        condition: "",
+        description: ""
+      };
     }
     log.message(
       "[%s] Deadlettering the message with id '%s'.",
--- a/lib/serviceBusMessage.ts
+++ b/lib/serviceBusMessage.ts
@@ -936,17 +936,14 @@ export class ServiceBusMessage implements ReceivedMessage {
    * rejecting the message.
    * @returns Promise<void>
    */
-  async deadLetter(options?: DeadLetterOptions): Promise<void> {
+  async deadLetter(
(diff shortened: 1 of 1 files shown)


lib/serviceBusMessage.ts
```diff
@@ -943,6 +943,11 @@ export class ServiceBusMessage implements ReceivedMessage {
         condition: options.deadletterReason,
         description: options.deadLetterErrorDescription
       };
+    } else {
```
AlexGhiondea: This could work, but I think we might be able to define the method like this instead:

```typescript
async deadLetter(options: DeadLetterOptions = {deadletterReason:"", deadLetterErrorDescription:"" }): Promise<void> {
```

If we do that, then if the options property is not specified it will be the empty strings (or some other values that we decide on).

/cc @amarzavery @ramya-rao-a

ShivangiReja: Result for both would be the same but your's is more succinct.

lib/serviceBusMessage.ts
```diff
@@ -936,7 +936,9 @@ export class ServiceBusMessage implements ReceivedMessage {
    * rejecting the message.
    * @returns Promise<void>
    */
-  async deadLetter(options?: DeadLetterOptions): Promise<void> {
+  async deadLetter(
+    options: DeadLetterOptions = { deadletterReason: "", deadLetterErrorDescription: "" }
+  ): Promise<void> {
     let error: AmqpError = {};
     if (options) {
```
AlexGhiondea: With the new method signature I think we can further simplify this code. Options will always be non-empty now, so I think we should always create the error object from the options argument. Depending on what is passed in, we are going to either put an empty string into the error object or actual non-empty strings that the user has specified.

lib/serviceBusMessage.ts
```diff
@@ -936,14 +936,14 @@ export class ServiceBusMessage implements ReceivedMessage {
    * rejecting the message.
    * @returns Promise<void>
    */
-  async deadLetter(options?: DeadLetterOptions): Promise<void> {
+  async deadLetter(
+    options: DeadLetterOptions = { deadletterReason: "", deadLetterErrorDescription: "" }
```
amarzavery: I wouldn't recommend using the default parameter syntax. I think when we compile it to ES6 our base target and ES5 for browsers, default parameter syntax is not transpiled correctly. Hence, it is better to avoid this syntax and set the default value inside the function if the parameter is null | undefined.

lib/serviceBusMessage.ts
```diff
@@ -936,14 +936,14 @@ export class ServiceBusMessage implements ReceivedMessage {
...
+  ): Promise<void> {
     let error: AmqpError = {};
-    if (options) {
-      error = {
-        condition: options.deadletterReason,
-        description: options.deadLetterErrorDescription
-      };
-    }
+    error = {
```
AlexGhiondea: Last nit :).

I think you can inline the definition of error and the setting of the properties:

```typescript
let error: AmqpError = {
  condition: options.deadletterReason,
  description: options.deadLetterErrorDescription
};
```

lib/serviceBusMessage.ts
AlexGhiondea: Interesting! I wasn't aware of this limitation. 

All right then -- back to the first implementation :). Sorry for the distraction @ShivangiReja

ShivangiReja: No problem, I will change this syntax.

Issue #109, opened 2018-12-19
Implement DeadletterQueue sample


AlexGhiondea:
https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/DeadletterQueue

ramya-rao-a:
Done with #128

PR #120, opened 2018-12-21
Add sample code for DeadLetterQueue usage

ramya0820:
Issue #109 

## Add samples for demonstrating usage of DLQ
- movingMessagesToDLQ
- processMessagesInDLQ

Added more information in a README file and inline comments, similar to the .NET implementation.
@AlexGhiondea @ramya-rao-a @ShivangiReja


AlexGhiondea:
Great to see the first PR out!

examples/deadletterqueue/README.md
```diff
@@ -0,0 +1,40 @@
+# Dead-Letter Queues
```
AlexGhiondea: Couple of comments:

- All our samples are in the examples\samples folder -- we should move the samples in the same folder.
- I would move the context from the readme file into the actual samples and remove the readme.md file

examples/deadletterqueue/movingMessagesToDLQ.ts
```diff
@@ -0,0 +1,192 @@
...
+  Namespace,
+  QueueClient
+} from "../../lib";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const queuePath = process.env.QUEUE_NAME || "";
+const numberOfMessages: number = parseInt(process.env.MESSAGE_COUNT || "1");
+const receiveClientTimeout: number = 10000;
```
AlexGhiondea: You don't need to specify `: number`

I would also make it clear what unit of time this is using (I assume this is miliseconds)

examples/deadletterqueue/movingMessagesToDLQ.ts
```diff
@@ -0,0 +1,192 @@
...
+  const onMessageHandler: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    await brokeredMessage.complete();
+  };
+
+  const receiverHandler = await client.receive(onMessageHandler, onError, { autoComplete: false });
+  await delay(receiveClientTimeout);
+  await receiverHandler.stop();
+}
+
+main();
```
AlexGhiondea: We should use the same pattern as the other samples for this.

Also, as main() returns a `Promise` we need to make sure that we get the result of the promise -- look at the other samples for an example of how to do that.

examples/deadletterqueue/movingMessagesToDLQ.ts
```diff
@@ -0,0 +1,192 @@
...
+  } catch (err) {
+    console.log(">>>>> Error in running sample scenarios: ", err);
+  } finally {
+    console.log("\n >>>> Calling close....");
+    ns.close();
+  }
+}
+
+// Queue clean up util
+async function purgeAllQueues(): Promise<void> {
```
AlexGhiondea: In general I think having multiple methods to split up the work is good, however in this case I don't think it is worth having this separate method. We can simply inline the code for it directly under the comment.

That way the sample is simpler and easier to follow.

examples/deadletterqueue/movingMessagesToDLQ.ts
```diff
@@ -0,0 +1,192 @@
...
+  await purgeQueue(queuePath + "/$DeadLetterQueue");
+}
+
+async function purgeQueue(inputQueuePath: string): Promise<void> {
+  // Process messages from queue, by invoking .deadletter() on the brokered message
+  await processMessageQueue(queuePurgingProcessor, inputQueuePath);
+  console.log(">>>> Cleaning up queue: ", inputQueuePath);
+}
+
+async function runSimpleDeadLetterScenario(): Promise<void> {
```
AlexGhiondea: I think we can also simplify the sample code here -- we can take the code that is inside these methods and put them inline at the call site.

For a sample I think having too many levels of indirection is confusing -- let's keep them as simple as possible to make them easier to follow and understand.

examples/deadletterqueue/movingMessagesToDLQ.ts
```diff
@@ -0,0 +1,192 @@
...
+  // Prepare given queue by sending few messages
+  await sendMessages();
+
+  // Process messages from queue, by invoking a faulty processor that doesn't successfully process
+  console.log("\n Setup queues, now running scenario C ... ");
+  await processMessageQueue(faultyProcessor, queuePath);
+}
+
+// Helper to send few sample messages to given queue
+async function sendMessages(): Promise<void> {
```
AlexGhiondea: We have a similar method in other samples (that one sets the TTL on the message as well as make the messages be distinct). I suggest you take a look at that method and see if we can re-use it here.

examples/deadletterqueue/movingMessagesToDLQ.ts
```diff
@@ -0,0 +1,192 @@
...
+  };
+
+  let i: number;
+  for (i = 0; i < numberOfMessages; i++) {
+    await client.send(message);
+  }
+}
+
+// Helper to process messages by delegating actual processing to a supplied custom messageProcessor
+async function processMessageQueue(messageProcessor: Function, path: string): Promise<void> {
```
AlexGhiondea: Once you inline this method we need to make sure that the client is closed.

examples/deadletterqueue/movingMessagesToDLQ.ts
```diff
@@ -0,0 +1,192 @@
...
+
+// OnError handler for message receive
+const onError: OnError = (err: MessagingError | Error) => {
+  console.log(">>>>> Error occurred: ", err);
+};
+
+// OnMessage handler for Scenario A - immediate dead lettering of message on receive
+async function simpleDeadletterProcessor(client: QueueClient): Promise<void> {
+  const onMessageHandler: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(">>>>> Killing the message", brokeredMessage);
```
AlexGhiondea: Let's say `deadletter the message` instead

examples/deadletterqueue/movingMessagesToDLQ.ts
```diff
@@ -0,0 +1,192 @@
...
+  await delay(receiveClientTimeout);
+  await receiverHandler.stop();
+}
+
+// OnMessage handler for Scenario C - usage for handling reject/fail on message processing
+async function faultyProcessor(client: QueueClient): Promise<void> {
+  const onMessageHandler: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(">>>>> Rejecting the message: ", brokeredMessage);
+    try {
+      await Promise.reject("Green apples are not OK.");
```
AlexGhiondea: Is this used in order to trigger and exception?

examples/deadletterqueue/processMessagesInDLQ.ts
```diff
@@ -0,0 +1,75 @@
...
+} from "../../lib";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const queuePath = process.env.QUEUE_NAME || "";
+const deadLetterQueuePath = queuePath + "/$DeadLetterQueue";
+const receiveClientTimeout: number = 10000;
+console.log("str: ", str);
+console.log("path: ", queuePath);
```
AlexGhiondea: Here we didn't log the same messages.

examples/deadletterqueue/processMessagesInDLQ.ts
```diff
@@ -0,0 +1,75 @@
...
+    await processMessageQueue(deadLetterMessageProcessor, deadLetterQueuePath);
+  } catch (err) {
+    console.log(">>>>> Error occurred in running sampple: ", err);
+  } finally {
+    console.log("\n>>>> Calling close....");
+    ns.close();
+  }
+}
+
+async function processMessageQueue(messageProcessor: Function, path: string): Promise<void> {
```
AlexGhiondea: Let's inline this method into the caller.

examples/deadletterqueue/processMessagesInDLQ.ts
```diff
@@ -0,0 +1,75 @@
...
+
+// This function takes in the message to inspect and reprocess it as needed.
+async function fixAndResendMessage(oldMessage: ServiceBusMessage): Promise<void> {
+  // Inspect given message and repair it
+  const repairedMessage = oldMessage.clone();
+  repairedMessage.body = { name: "Apple", color: "Red", qty: 5 };
+
+  // Send repaired message back to the current queue
+  const client = ns.createQueueClient(queuePath);
+  await client.send(repairedMessage);
```
AlexGhiondea: We need to close the client.

examples/deadletterqueue/processMessagesInDLQ.ts
```diff
@@ -0,0 +1,75 @@
...
+    await brokeredMessage.complete();
+  };
+
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+
+  const receiverHandler = await client.receive(onMessageHandler, onError, { autoComplete: false });
+  await delay(receiveClientTimeout);
+  await receiverHandler.stop();
```
AlexGhiondea: We need to close the client (possibly in the caller method).

examples/deadletterqueue/README.md
ramya0820: Agree, addressed.

examples/deadletterqueue/movingMessagesToDLQ.ts
ramya0820: Updated.

examples/deadletterqueue/movingMessagesToDLQ.ts
ramya0820: Agree, updated to maintain uniformity.

examples/deadletterqueue/movingMessagesToDLQ.ts
ramya0820: Yes, updated code to throw it as an error.

examples/deadletterqueue/movingMessagesToDLQ.ts
ramya0820: Discussed offline, updated code to handle promise returned on main.

examples/deadletterqueue/processMessagesInDLQ.ts
ramya0820: Updated

examples/deadletterqueue/processMessagesInDLQ.ts
ramya0820: Updated

examples/deadletterqueue/processMessagesInDLQ.ts
ramya0820: Updated

Issue #60, opened 2018-12-11
Implement the PartitionedQueues sample


AlexGhiondea:
https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/PartitionedQueues

ramya-rao-a:
Done with #123

PR #123, opened 2018-12-25
Ported PartitionedQueues .NET sample to typescript

HarshaNalluru:
Ported the PartitionedQueues .NET sample to typescript.
Any comments ?
[Original .NET sample](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/PartitionedQueues)
Issue #60

### DIFF
--- a/examples/samples/partitionedQueues.ts
+++ b/examples/samples/partitionedQueues.ts
@@ -1,7 +1,7 @@
+// Enable partitions for the topics in the azure portal and then execute the sample
 import {
   delay,
   SendableMessageInfo,
-  QueueClient,
   ReceiveMode,
   generateUuid,
   Namespace,
@@ -19,46 +19,14 @@ const path = process.env.QUEUE_NAME || "";
 console.log(`str: ${str}`);
 console.log(`path: ${path}`);
 
-let ns: Namespace;
 async function main(): Promise<void> {
-  ns = Namespace.createFromConnectionString(str);
-  const client = ns.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
-
-  // populate the queue with messages
-  await sendMessages(client);
-
-  let rcvHandler: ReceiveHandler;
-
-  // retrieve messages from the queue
-  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
-    console.log(
-      ` \n### Received message:
-        ID - ${brokeredMessage.messageId},
-        messageBody - ${brokeredMessage.body ? brokeredMessage.body.toString() : undefined},
-        partitionKey - ${brokeredMessage.partitionKey},
-        label - ${brokeredMessage.label}`
-    );
-  };
-
-  const onError: OnError = (err: MessagingError | Error) => {
-    console.log(">>>>> Error occurred: ", err);
-  };
-
-  rcvHandler = client.receive(onMessage, onError);
-
-  // wait 5 seconds
-  await delay(5000);
-
-  console.log("Stopping the receiver");
-
-  await rcvHandler.stop();
-
-  console.log("Closing the client");
-
-  await client.close();
+  await sendMessages();
+  await receiveMessages();
 }
 
-async function sendMessages(queueClient: QueueClient): Promise<void> {
+async function sendMessages(): Promise<void> {
+  const nsSend = Namespace.createFromConnectionString(str);
+  const sendClient = nsSend.createQueueClient(path);
   const data = [
     { lastName: "Einstein", firstName: "Albert" },
     { lastName: "Heisenberg", firstName: "Werner" },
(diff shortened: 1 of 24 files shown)


examples/samples/partitionedQueues.ts
```diff
@@ -0,0 +1,97 @@
...
+    { lastName: "Hawking", firstName: "Steven" },
+    { lastName: "Newton", firstName: "Isaac" },
+    { lastName: "Bohr", firstName: "Niels" },
+    { lastName: "Faraday", firstName: "Michael" },
+    { lastName: "Galilei", firstName: "Galileo" },
+    { lastName: "Kepler", firstName: "Johannes" },
+    { lastName: "Kopernikus", firstName: "Nikolaus" }
+  ];
+
+  for (let j = 0; j < 3; j++) {
```
ramya-rao-a: Any idea why we need this outer loop?

examples/samples/partitionedQueues.ts
```diff
@@ -0,0 +1,97 @@
...
+  let rcvHandler: ReceiveHandler;
+
+  // retrieve messages from the queue
+  const onMessage: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
+    console.log(
+      ` \n### Received message:
+        ID - ${brokeredMessage.messageId},
+        messageBody - ${brokeredMessage.body ? brokeredMessage.body.toString() : undefined},
+        partitionKey - ${brokeredMessage.partitionKey},
+        label - ${brokeredMessage.label}`
```
ramya-rao-a: The .Net sample talks about how the sequence number is affected in partitioned queues. Any idea what that is about and how we are showing the same here?

examples/samples/partitionedQueues.ts
```diff
@@ -0,0 +1,97 @@
...
+
+let ns: Namespace;
+async function main(): Promise<void> {
+  ns = Namespace.createFromConnectionString(str);
+  const client = ns.createQueueClient(path, { receiveMode: ReceiveMode.peekLock });
+
+  // populate the queue with messages
+  await sendMessages(client);
+
+  let rcvHandler: ReceiveHandler;
```
ramya-rao-a: All other samples have the receiving part of the sample in a separate function. Can we do the same here to maintain consistency?

examples/samples/partitionedQueues.ts
HarshaNalluru: SequenceNumber value is a unique 64-bit integer assigne4d  by ServiceBus. For partitioned queues, the topmost 16 bits reflect the partition id. Sequence numbers roll over to zero when the 48/64-bit range is exhausted. The usual the ascending SequenceNumber characteristics is no longer guaranteed as they depend on partitionId.

I think I should print the SequenceNumber as well to clear the confusion.

examples/samples/partitionedQueues.ts
HarshaNalluru: This would help to show that the messages with same partitionId will go to the same fragment. 
As I mentioned in the below comment(https://github.com/Azure/azure-service-bus-node/pull/123#discussion_r244848283), it makes more sense when we print the SequenceNumber.

examples/samples/partitionedQueues.ts
HarshaNalluru: Not all the samples. I think I followed this convention from SendersReceiversWithQueues sample and kept the receiving part in the main module. 
Shall I create a separate function ?

ramya-rao-a: I meant all other samples in this repo

HarshaNalluru: Okay, I'll change this :)

examples/samples/partitionedQueues.ts
HarshaNalluru: In the latest commit, updated the print statement to print `SequenceNumber` in the received message.
And at the beginning of the code, added a comment to enable partitions in the azure portal for the topic.

This would clear the confusion ?

examples/samples/partitionedQueues.ts
ramya-rao-a: Currently, the first 2 letters of the name are being used as the partition key. If you use just the first letter, then you can skip the loop. The multiple names starting with the same letter will then have the same partitionId and will go to the same fragment.

To make the sample even more simpler, you can also change your data to be such that few of the names share the same last name. Then you can use the last name as the partition key.

examples/samples/partitionedQueues.ts
ramya-rao-a: Sure. Also add a comment on what the user is expected to see

examples/samples/partitionedQueues.ts
HarshaNalluru: I agree, this would significantly reduce the number of messages to send and would simplify the sample.

HarshaNalluru:
Can I add a link to the [Docs](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning) in the sample(as a comment) ?

Many links inside the docs point to .NET modules. Suggest something @ramya-rao-a

ramya-rao-a:
Yes, you can add links to the docs.

Eventually the docs will point to appropriate modules

Issue #65, opened 2018-12-11
Implement Sessions sample


AlexGhiondea:
https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/Sessions

PR #124, opened 2018-12-27
Add Session sample and resolve linter errors.

ShivangiReja:
* Added Session sample #65
* Resolved all the linter errors.
* Change topicGettingstarted sample(follow the same pattern as others).

### DIFF
--- a/examples/samples/session.ts
+++ b/examples/samples/session.ts
@@ -72,8 +72,7 @@ async function receiveMessages(queueClient: QueueClient): Promise<void> {
     console.log(">>>>> Error occurred: ", err);
   };
   queueClient.receiveMessgesFromSessions(onMessage, onError);
-  await delay(20000);
-  await queueClient.close();
+  await delay(10000);
 }
 
 main()
--- a/examples/samples/session.ts
+++ b/examples/samples/session.ts
@@ -66,7 +66,6 @@ async function receiveMessages(queueClient: QueueClient): Promise<void> {
     console.log(
       `Message received: ${brokeredMessage.body} SessionId : ${brokeredMessage.sessionId}`
     );
-    await brokeredMessage.complete();
   };
   const onError: OnError = (err: MessagingError | Error) => {
     console.log(">>>>> Error occurred: ", err);
@@ -76,10 +75,10 @@ async function receiveMessages(queueClient: QueueClient): Promise<void> {
 }
 
 main()
+  .catch((err) => {
+    console.log("error: ", err);
+  })
   .then(() => {
     console.log(">>>> Calling close....");
     return ns.close();
-  })
-  .catch((err) => {
-    console.log("error: ", err);
   });
--- a/examples/samples/session.ts
+++ b/examples/samples/session.ts
@@ -32,6 +32,7 @@ async function main(): Promise<void> {
   await receiveMessages(client);
 
   await client.close();
+  ns.close();
 }
 
 async function sendMessages(queueClient: QueueClient, sessionId: string): Promise<void> {
@@ -75,10 +76,10 @@ async function receiveMessages(queueClient: QueueClient): Promise<void> {
 }
 
 main()
-  .catch((err) => {
-    console.log("error: ", err);
-  })
   .then(() => {
     console.log(">>>> Calling close....");
-    return ns.close();
+  })
+  .catch((err) => {
+    console.log("error: ", err);
+    ns.close();
   });
--- a/examples/samples/session.ts
+++ b/examples/samples/session.ts
@@ -32,7 +32,7 @@ async function main(): Promise<void> {
   await receiveMessages(client);
 
   await client.close();
-  ns.close();
+  await ns.close();
 }
 
 async function sendMessages(queueClient: QueueClient, sessionId: string): Promise<void> {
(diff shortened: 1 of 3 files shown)


examples/samples/session.ts
```diff
@@ -0,0 +1,85 @@
...
+    console.log(">>>>> Error occurred: ", err);
+  };
+  queueClient.receiveMessgesFromSessions(onMessage, onError);
+  await delay(10000);
+}
+
+main()
+  .then(() => {
+    console.log(">>>> Calling close....");
+    return ns.close();
```
HarshaNalluru: @ramya-rao-a suggested adding a try-catch in main and invoke `ns.close()` once in the finally block - https://github.com/Azure/azure-service-bus-node/pull/127/files

HarshaNalluru: @ramya-rao-a suggested adding a try-catch in main and invoke ns.close() once in the finally block - https://github.com/Azure/azure-service-bus-node/pull/127/files

ShivangiReja: @ramya0820 suggested this comment, @ramya-rao-a is on vacations.

@HarshaNalluru  @ramya0820 I was following the same pattern as old samples. I think before changing/merging anything we should wait for @ramya-rao-a suggestion.
I don't see any additional benefit of adding extra try-catch-finally block in a separate method. I agree that we should close namespace/queueClient etc before exiting but instead of adding new method , we can write 
```
main()
  .catch((err) => {
    console.log("error: ", err);
  })
  .then(() => {
    console.log(">>>> Calling close....");
    return ns.close();
  });
```
It would give us the same result as try, catch and finally.
but still I would wait for @ramya-rao-a @AlexGhiondea suggestions on this.

HarshaNalluru: It looks cleaner this way. Sure, let's wait for @ramya-rao-a.

ramya0820: Chaining .then() is definitely not the same as using a finally block (such as on force quitting programs).

Actually, it looks like we can use .finally() on a promise
https://blog.mariusschulz.com/2018/11/27/using-promise-prototype-finally-in-typescript

I think this would be our best bet.

examples/samples/session.ts
```diff
@@ -0,0 +1,85 @@
...
+
+async function receiveMessages(queueClient: QueueClient): Promise<void> {
+  const onMessage: OnSessionMessage = async (
+    messageSession: MessageSession,
+    brokeredMessage: ServiceBusMessage
+  ) => {
+    console.log(
+      `Message received: ${brokeredMessage.body} SessionId : ${brokeredMessage.sessionId}`
+    );
+    await brokeredMessage.complete();
```
ramya-rao-a: Since autoComplete is enabled by default, do we need to manually complete the message here?

examples/samples/session.ts
ramya-rao-a: I am missing something here... Can someone point me to the original comment?

examples/samples/session.ts
ShivangiReja: ohh yes ,I forgot to remove `await brokeredMessage.complete();`

examples/samples/session.ts
```diff
@@ -0,0 +1,85 @@
...
+  };
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log(">>>>> Error occurred: ", err);
+  };
+  queueClient.receiveMessgesFromSessions(onMessage, onError);
```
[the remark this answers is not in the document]
ShivangiReja: No, we don't have the specific sample for getting messages from specific session but I can change this sample to call receive method for the specific session or we can make one more sample for getting messages from a specific session.
@ramya-rao-a what do you suggest ?

examples/samples/session.ts
ShivangiReja: In this way, we have to write `ns.close()` twice ?
```
main()
  .then(() => {
    console.log(">>>> Calling close....");
    return ns.close();
  })
  .catch((err) => {
    console.log("error: ", err);
    ns.close();
  });`

Comment (ramya-rao-a) [reply examples/samples/session.ts]: Yes, but the first one would be inside the `main()` call right after we close the client which would seem more natural.

As a user, I am done working with the client, so I call close on it. I am done with the namespace, so I call close on it. One after the other.

The `main().catch()` would be for unhandled error scenarios

File: examples/samples/session.ts (hunk 3 above)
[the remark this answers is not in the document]
Comment (ramya-rao-a) [reply examples/samples/session.ts]: I believe it would be useful to show all both ways of receiving
- Receive from a particular session (pick one of the sessions)
- Receive from all sessions (this will only get messages from rest of the sessions)

This way, the user can decide which way they want to go based on their use case

But this also goes against the "have minimal code in samples" argument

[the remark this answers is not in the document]
Comment (ShivangiReja) [reply examples/samples/session.ts]: we can make one more sample specific for getting messages from a specific session, in this way we will have both type of examples in simpler way for the users.

File: examples/samples/session.ts [hunk 4]
```diff
@@ -32,6 +32,7 @@ async function main(): Promise<void> {
   await receiveMessages(client);
 
   await client.close();
+  ns.close();
```
ramya-rao-a: Missing `await`

Issue #125, opened 2018-12-27
Implement sample to demo TimeToLive usage


ramya0820:

ramya0820:
CC-ing team for review - @AlexGhiondea @ramya-rao-a @ShivangiReja @HarshaNalluru

ramya-rao-a:
Done with #127

PR #127, opened 2018-12-28
Add sample to demonstrate TimeToLive property usage

ramya0820:
Issue #125

## Description

Added sample code to demonstrate TimeToLive property usage

- Issue #125 

@AlexGhiondea @ShivangiReja @HarshaNalluru

### DIFF
--- a/examples/samples/timeToLive.ts
+++ b/examples/samples/timeToLive.ts
@@ -64,9 +64,8 @@ async function sendMessage(): Promise<void> {
 }
 
 async function receiveMessageFromQueue(givenQueuePath: string): Promise<void> {
-  // Process messages from queue, by invoking .deadletter() on the brokered message
   const client = ns.createQueueClient(givenQueuePath, { receiveMode: ReceiveMode.peekLock });
-  console.log("\n>>>>> Receiving message from queue", givenQueuePath);
+
   const onMessageHandler: OnMessage = async (brokeredMessage: ServiceBusMessage) => {
     console.log(">>>>> Received message with data", brokeredMessage.body);
     await brokeredMessage.complete();
@@ -76,10 +75,12 @@ async function receiveMessageFromQueue(givenQueuePath: string): Promise<void> {
     console.log(">>>>> Error occurred: ", err);
   };
 
+  console.log("\n>>>>> Receiving message from queue", givenQueuePath);
   const receiverHandler = await client.receive(onMessageHandler, onError, { autoComplete: false });
   await delay(receiveClientTimeoutInMilliseconds);
   await receiverHandler.stop();
   console.log(">>>>> Done receiving message from queue", givenQueuePath);
+
   await client.close();
 }
 
--- a/examples/samples/timeToLive.ts
+++ b/examples/samples/timeToLive.ts
@@ -25,9 +25,6 @@ let ns: Namespace;
 /*
   This sample demonstrates how TimeToLive property works. For more information, refer to
   https://docs.microsoft.com/en-us/azure/service-bus-messaging/message-expiration
-
-  On running this sample, you should immediately see a Pineapple Icecream message in the main queue.
-  After 30 seconds, you should see the message moved to DLQ with appropriate reason being set.
 */
 async function main(): Promise<void> {
   ns = Namespace.createFromConnectionString(str);
@@ -45,9 +42,9 @@ async function main(): Promise<void> {
   // This will contain the message details that was sent, implying the message is moved to DLQ
   await receiveMessageFromQueue(deadLetterQueuePath);
 
-  console.log(">>>> Calling close....");
+  console.log("\n>>>> Calling close....");
   await ns.close();
-  console.log(">>>> sample Done!!!!");
+  console.log("\n>>>> sample Done!!!!");
 }
 
 async function sendMessage(): Promise<void> {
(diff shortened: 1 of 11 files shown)


examples/samples/timeToLive.ts
```diff
@@ -0,0 +1,105 @@
...
+  Namespace,
+  SendableMessageInfo,
+  generateUuid
+} from "../../lib";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+const str = process.env.SERVICEBUS_CONNECTION_STRING || "";
+const queuePath = process.env.QUEUE_NAME || "";
+const deadLetterQueuePath = queuePath + "/$DeadLetterQueue";
```
ramya-rao-a: There is a `Namespace.getDeadLetterQueuePathForQueue()` function you can use to get the deadletter queue path

examples/samples/timeToLive.ts
```diff
@@ -0,0 +1,105 @@
...
+    label: "Recipe",
+    timeToLive: 30 * 1000, // 30 seconds
+    messageId: generateUuid()
+  };
+  await client.send(message);
+  await client.close();
+}
+
+// Task for processing the expired messages after they land in the Dead Letter Queue
+async function PickUpAndFixDeadletters(): Promise<void> {
```
ramya-rao-a: As a user, I would expect the demonstration of time to live feature would be to
- let the message expire
- prove that normal way of receiving messages doesnt receive the messages, well because they expired
- prove that the expired messages are in the dead letter queue 

The whole concept of fixing these messages and re-submitting feels un-necessary for this sample. It may have a place in the sample for DeadLetter queue, but not for Time To Live 

Thoughts?

cc @ShivangiReja , @ramya0820 , @AlexGhiondea, @HarshaNalluru

ShivangiReja: Yes, I think for this sample fixing and resubmitting is not necessary because we trying to make these samples as simple as possible for the users, so there is no need to add extra functionality.

examples/samples/timeToLive.ts
ramya0820: Got it, updated this and DLQ samples too.

examples/samples/timeToLive.ts
```diff
@@ -0,0 +1,105 @@
...
+    // Expire the messages
+    await delay(60000);
+
+    // Process expired messages from the Dead Letter Queue
+    await PickUpAndFixDeadletters();
+  } catch (err) {
+    console.log(">>>>> Error occurred in running sampple: ", err);
+  } finally {
+    console.log("\n>>>> Calling close....");
+    ns.close();
```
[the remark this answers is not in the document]
ramya0820: Oh, the finally was wrapping over all of the code in main so I'm a little confused why it would miss errors thrown by any of the functions invoked from within it.
But yes, agree with dropping it for purpose of samples - have updated code in this and others..

examples/samples/timeToLive.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+// Task to send messages to given queue
+async function SendMessagesAsync(): Promise<void> {
```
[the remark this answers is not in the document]
ramya0820: Agree, I should have caught this in first place!

examples/samples/timeToLive.ts
```diff
@@ -0,0 +1,94 @@
...
+  // Attempt to receive message from main queue
+  // This will not show the message details that was sent, implying message got expired and removed
+  await receiveMessageFromQueue(queuePath);
+
+  // Attempt to receive message from DLQ
+  // This will contain the message details that was sent, implying the message is moved to DLQ
+  await receiveMessageFromQueue(deadLetterQueuePath);
+
+  console.log(">>>> Calling close....");
+  ns.close();
```
ramya-rao-a: await this

examples/samples/timeToLive.ts
```diff
@@ -0,0 +1,94 @@
...
+  const receiverHandler = await client.receive(onMessageHandler, onError, { autoComplete: false });
+  await delay(receiveClientTimeoutInMilliseconds);
+  await receiverHandler.stop();
+  await client.close();
+}
+
+main()
+  .catch((err) => {
+    console.log(">>>>> Error occurred: ", err);
+    ns.close();
```
ramya-rao-a: If you have a `then` after this, then `ns.close()` need to be returned

examples/samples/timeToLive.ts
```diff
@@ -0,0 +1,89 @@
...
+    timeToLive: 30 * 1000, // 30 seconds
+    messageId: generateUuid()
+  };
+  await client.send(message);
+  console.log(">>>>> Sent message with data", recipeData);
+  await client.close();
+}
+
+async function receiveMessageFromQueue(givenQueuePath: string): Promise<void> {
+  // Process messages from queue, by invoking .deadletter() on the brokered message
```
ramya-rao-a: Misplaced comment?

examples/samples/timeToLive.ts
```diff
@@ -0,0 +1,89 @@
...
+  const client = ns.createQueueClient(givenQueuePath, { receiveMode: ReceiveMode.peekLock });
+  console.log("\n>>>>> Receiving message from queue", givenQueuePath);
```
ramya-rao-a: For ease of readability, I would move this console.log to just before `client.receive`

Issue #99, opened 2018-12-19
Implement tests for Send


AlexGhiondea:
- [x] Simple Send and ReceiveBatch in Queue/Topic/Subscription
- [x] SendBatch and ReceiveBatch in Queue/Topic/Subscription
- [x] ScheduleMessage, ScheduleMessages using Queue/Topics/Subscription.

ramya-rao-a:
@ShivangiReja I am adding the test item for cancelling a scheduled message in the above list. So, re-opening this issue

AlexGhiondea:
@ramya-rao-a would it make sense to create a new issue for this?

ramya-rao-a:
I am ok either way. Will create a new issue

PR #129, opened 2018-12-28
Add testcases for schedule messages and resolve existing linter errors.

ShivangiReja:
* Added test cases for schedule messages.
* Resolved existing linter errors. #99

### DIFF
--- a/examples/samples/session.ts
+++ b/examples/samples/session.ts
@@ -1,0 +1,85 @@
+import {
+  generateUuid,
+  QueueClient,
+  SendableMessageInfo,
+  OnSessionMessage,
+  OnError,
+  MessagingError,
+  delay,
+  ServiceBusMessage,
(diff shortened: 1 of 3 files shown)


test/sendReceive.spec.ts
```diff
@@ -30,7 +30,7 @@ const testMessages: SendableMessageInfo[] = [
   }
 ];
 
-function testReceivedMessages(receivedMsgs: ServiceBusMessage[]) {
+async function testReceivedMessages(receivedMsgs: ServiceBusMessage[]): Promise<void> {
```
ramya-rao-a: Everything inside `testReceivedMessages` is synchronous. Why make it `async`?

ShivangiReja: I was resolving linter errors. I will remove async.

PR #140, opened 2019-01-03
Detail that the QUEUE_NAME should point to a queue that doesn't have sessions enabled

AlexGhiondea:


test/README.md
```diff
@@ -4,7 +4,7 @@
 - Have a file by the name `.env` in the root folder of this project.
 - This file should contain the below environment variables pointing to the various Service Bus resources
     - SERVICEBUS_CONNECTION_STRING
-    - QUEUE_NAME
+    - QUEUE_NAME (without sessions enabled)
```
ramya-rao-a: well... we eventually will have tests for sessions as well
So I would rather add a new variable called `SESSION_ENABLED_QUEUE_NAME`

AlexGhiondea: I agree -- But I think we need to call out that this queue doesn't have sessions enabled. If you do, the tests will fail.

PR #141, opened 2019-01-03
Resolved : Deadletter related properties error.

ShivangiReja:

### DIFF
--- a/lib/serviceBusMessage.ts
+++ b/lib/serviceBusMessage.ts
@@ -949,17 +949,13 @@ export class ServiceBusMessage implements ReceivedMessage {
    * @returns Promise<void>
    */
   async deadLetter(options?: DeadLetterOptions): Promise<void> {
-    let error: AmqpError = {
-      condition: "",
-      description: ""
+    const error: AmqpError = {
+      condition: Constants.deadLetterName
     };
     if (options) {
-      error = {
-        condition: Constants.deadLetterName,
-        info: {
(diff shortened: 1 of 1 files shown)


lib/serviceBusMessage.ts
```diff
@@ -955,8 +955,11 @@ export class ServiceBusMessage implements ReceivedMessage {
     };
     if (options) {
       error = {
-        condition: options.deadletterReason,
-        description: options.deadLetterErrorDescription
+        condition: Constants.deadLetterName,
```
ramya-rao-a: I would suggest to set this condition in line 953 and remove the description on line 954

AlexGhiondea: Great catch!

lib/serviceBusMessage.ts
```diff
@@ -949,14 +949,13 @@ export class ServiceBusMessage implements ReceivedMessage {
...
+    const error: AmqpError = {
+      condition: Constants.deadLetterName
     };
     if (options) {
-      error = {
-        condition: options.deadletterReason,
-        description: options.deadLetterErrorDescription
+      error.info = {
+        DeadLetterReason: options.deadletterReason,
+        DeadLetterErrorDescription: options.deadLetterErrorDescription
```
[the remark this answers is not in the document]
ramya-rao-a: AMQP requires the pascal casing.
camelCasing is the norm for nodejs. 
So I would keep this as is.

Issue #64, opened 2018-12-11
Implement SessionState sample


AlexGhiondea:
https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/SessionState

ramya-rao-a:
Done with #148

Issue #144, opened 2019-01-04
Abandon or Deadlettering a message received from receiveBatch takes 20 seconds


ramya-rao-a:
**Describe the bug**

Receive a message by using `receiveBatch()`
Run `await message.abandon()` or `await message.deadLetter()`

This takes about 20 seconds.

Completing a message doesnt take this time

ramya-rao-a:
I can confirm that I no longer see this happening.
Therefore closing this issue

PR #148, opened 2019-01-04
Implement sample for SessionState

ramya0820:
Sample for session state using Shopping Cart scenario as discussed.

## Description
Includes bug fix for getState() to return the state and not void

- #64 

@AlexGhiondea @ramya-rao-a 
@ShivangiReja @HarshaNalluru

### DIFF
--- a/examples/samples/sessionState.ts
+++ b/examples/samples/sessionState.ts
@@ -12,7 +12,7 @@ let ns: Namespace;
 
 /*
   This sample demonstrates usage of SessionState and how to use it.
-  Ref - <link>
+  Ref - https://docs.microsoft.com/en-us/azure/service-bus-messaging/message-sessions
 
   We take for example the context of an online shopping app and see how we can use Session State
   to implement the maintaining of shopping cart information completely on server side i.e.,
--- a/examples/samples/sessionState.ts
+++ b/examples/samples/sessionState.ts
@@ -16,7 +16,7 @@ let ns: Namespace;
 
   We take for example the context of an online shopping app and see how we can use Session State
   to implement the maintaining of shopping cart information completely on server side i.e.,
-  remembering the user's shopped items even if they leave the site and return later.
+  remembering the users' shopped items even if they leave the site and return later.
 
   The scenario in sample walks through user activity of two customers Alice and Bob.
   Alice shops for 3 items and checks out, whereas Bob adds 3 items and leaves without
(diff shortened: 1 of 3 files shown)


examples/samples/sessionState.ts
```diff
@@ -0,0 +1,146 @@
...
+/*
+  This sample demonstrates usage of SessionState.
+  Ref - https://docs.microsoft.com/en-us/azure/service-bus-messaging/message-sessions
+
+  We take for example the context of an online shopping app and see how we can use Session State
+  to implement the maintaining of shopping cart information completely on server side i.e.,
+  remembering the user's shopped items even if they leave the site and return later.
+
+  The scenario in sample walks through user activity of two customers Alice and Bob.
+  Alice shops for 3 items and checks out, whereas Bob adds 3 items and leaves without
```
ramya-rao-a: "Bob adds 3 items and" should either be "Bob shops for 3 items and". Otherwise it looks like Alice and Bob did 2 different activities.

Or since you later refer to items in cart, may be both Alice and Bob should "add 3 items and. .."

examples/samples/sessionState.ts
```diff
@@ -0,0 +1,146 @@
...
+  // Get list of items
+  console.log(`\nItems in cart for ${sessionId}: ${sessionState}\n`);
+
+  await messageSession.close();
+  await client.close();
+}
+
+async function sendMessagesForSession(shoppingEvents: any[], sessionId: string): Promise<void> {
+  const client = ns.createQueueClient(userEventsQueue, {
+    receiveMode: ReceiveMode.peekLock
```
ramya-rao-a: This is the default mode, we can skip passing the `options` altogether. This applies to the other place where we use queueClient as well

examples/samples/sessionState.ts
```diff
@@ -0,0 +1,146 @@
...
+    await client.send(message);
+  }
+  await client.close();
+}
+
+async function processMessageFromSession(sessionId: string): Promise<void> {
+  const client = ns.createQueueClient(userEventsQueue, { receiveMode: ReceiveMode.peekLock });
+  const messageSession = await client.acceptSession({ sessionId: sessionId });
+
+  // TODO: Fix bug #144
```
ramya-rao-a: What does #144 has to do here? We are not abandoning or deadlettering the message, are we?

examples/samples/sessionState.ts
```diff
@@ -0,0 +1,146 @@
...
+  await getSessionState("bob");
+}
+
+async function getSessionState(sessionId: string): Promise<void> {
+  const client = ns.createQueueClient(userEventsQueue, { receiveMode: ReceiveMode.peekLock });
+  const messageSession = await client.acceptSession({ sessionId: sessionId });
+
+  const sessionState = await messageSession.getState();
+  // Get list of items
+  console.log(`\nItems in cart for ${sessionId}: ${sessionState}\n`);
```
ramya-rao-a: Perhaps, a different message can be logged when no items are there in the cart?

Would you consider differentiating the cases of no items in the cart because none were added vs because user checked out?

PR #150, opened 2019-01-07
Implement some tests for ReceiveBatch in PeekLock mode

ShivangiReja:


test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
 
     await receivedMsgs[0].complete();
   });
+
+  it("Abandoning a message 10 times, you get it no more. You can then only get it from the dead letter queue", async function(): Promise<
```
ramya-rao-a: I wrote the above line in the github issue so that the text was self explainatory to whoever was writing the tests. We can improve it here. How about this:

> Message abandoned more than maxDeliveryCount goes to dead letter queue

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+      should.equal(receivedMsgs[0].messageId, testMessages[0].messageId);
+      should.equal(receivedMsgs[0].deliveryCount, abandonMsgCount);
+      abandonMsgCount++;
+
+      await receivedMsgs[0].abandon();
+    }
+
+    await testPeekMsgsLength(queueClient, 0);
+
+    const deadLetterQueuePath = process.env.QUEUE_NAME + "/$DeadLetterQueue";
```
ramya-rao-a: Let's use the helper method `getDeadLetterQueuePathForQueue` from Namespace here

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    const client = namespace.createQueueClient(deadLetterQueuePath);
```
ramya-rao-a: Rename `client` to `deadletterQueueClient` for ease of readability?

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+      should.equal(receivedMsgs.length, 1);
+      should.equal(receivedMsgs[0].messageId, testMessages[0].messageId);
+      should.equal(receivedMsgs[0].deliveryCount, abandonMsgCount);
+      abandonMsgCount++;
+
+      await receivedMsgs[0].abandon();
+    }
+
+    await testPeekMsgsLength(queueClient, 0);
```
ramya-rao-a: We should probably close the `queueClient` after this

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+    const deadLetterQueuePath = process.env.QUEUE_NAME + "/$DeadLetterQueue";
+    const client = namespace.createQueueClient(deadLetterQueuePath);
+    receivedMsgs = await client.receiveBatch(1);
```
ramya-rao-a: Let's use a different variable here for easier differentiation between the normal msgs and the dead letter ones. Say `deadLetterMsgs`?

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
   });
+
+  it("Abandoning a message 10 times, you get it no more. You can then only get it from the dead letter queue", async function(): Promise<
+    void
+  > {
+    await queueClient.send(testMessages[0]);
+    let abandonMsgCount = 0;
+    let receivedMsgs;
+
+    while (abandonMsgCount < 10) {
```
ramya-rao-a: Let's have this 10 as a constant (`maxDeliveryCount`) and define it outside in the describe() method.

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+      await receivedMsgs[0].abandon();
+    }
+
+    await testPeekMsgsLength(subscriptionClient, 0);
+
+    const deadLetterTopicPath =
+      process.env.TOPIC_NAME +
+      "/Subscriptions/" +
+      process.env.SUBSCRIPTION_NAME +
+      "/$DeadLetterQueue";
```
ramya-rao-a: Let's create a new helper method for this in namespace.ts file, just like we have `getDeadLetterQueuePathForQueue` for queues

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+      should.equal(receivedMsgs[0].messageId, testMessages[0].messageId);
+      should.equal(receivedMsgs[0].deliveryCount, abandonMsgCount);
+      abandonMsgCount++;
+
+      await receivedMsgs[0].abandon();
+    }
+
+    await testPeekMsgsLength(subscriptionClient, 0);
+
+    const deadLetterTopicPath =
```
ramya-rao-a: Should this be `deadLetterSubscriptionPath`?

I am guessing each subscription gets to have its own dead letter sub-subscription

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    should.equal(receivedMsgs.length, 1);
+    should.equal(receivedMsgs[0].body, testMessages[0].body);
+    should.equal(receivedMsgs[0].messageId, testMessages[0].messageId);
+
+    await receivedMsgs[0].complete();
+
+    await testPeekMsgsLength(clientSubscription, 0);
+  });
+
+  it("Deferring a message results in not getting the same message again from queue. The message is then gotten using receiveDefferedMessages", async function(): Promise<
```
ramya-rao-a: Better name would be `Receive deferred message from queue`

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+      await receivedMsgs[0].complete();
+
+      const message = await queueClient.receiveDeferredMessage(sequenceNumber);
+      if (message) {
+        should.equal(message.body, testMessages[0].body);
+        should.equal(message.messageId, testMessages[0].messageId);
+
+        await message.complete();
+      } else {
+        throw "The message that we are receving from the sequence number can not be null";
```
ramya-rao-a: We can simplify this to - > "No message received for sequence number"

test/batchReceiver.spec.ts
```diff
@@ -213,4 +213,157 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+  > {
+    await queueClient.sendBatch(testMessages);
+    const msgs = await queueClient.receiveBatch(1);
+
+    should.equal(Array.isArray(msgs), true);
+    should.equal(msgs.length, 1);
+    should.equal(msgs[0].body, testMessages[0].body);
+    should.equal(msgs[0].messageId, testMessages[0].messageId);
+
+    if (msgs[0].sequenceNumber) {
```
ramya-rao-a: Instead of nested if/else blocks, we can simplify this code by exiting early. For example:

```
if (!msgs[0].sequenceNumber) {
   // throw the error here
}

// receive msg from queue to test that it is not the same as the deferred msg
// use sequence number to receive the deferred msg

if (!message) {
   // throw error
}

// test message contents
```

test/batchReceiver.spec.ts
```diff
@@ -234,31 +235,32 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
 
     await testPeekMsgsLength(queueClient, 0);
 
-    const deadLetterQueuePath = process.env.QUEUE_NAME + "/$DeadLetterQueue";
-    const client = namespace.createQueueClient(deadLetterQueuePath);
-    receivedMsgs = await client.receiveBatch(1);
+    const deadLetterQueuePath = Namespace.getDeadLetterQueuePathForQueue(
+      process.env.QUEUE_NAME ? process.env.QUEUE_NAME : ""
```
ramya-rao-a: How about using `queueClient.name`? Same thing for subscription

test/batchReceiver.spec.ts
```diff
@@ -234,31 +235,32 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    should.equal(deadLetterMsgs[0].body, testMessages[0].body);
+    should.equal(deadLetterMsgs[0].messageId, testMessages[0].messageId);
 
-    await receivedMsgs[0].complete();
+    await deadLetterMsgs[0].complete();
 
-    await testPeekMsgsLength(client, 0);
-    await client.close();
+    await testPeekMsgsLength(deadletterQueueClient, 0);
+    await deadletterQueueClient.close();
```
ramya-rao-a: `namespace.close()` is called at the end of each test which will close all the queue/topic/subscription clients. So maybe we can skip this close()?

test/batchReceiver.spec.ts
ramya-rao-a: Do you get any errors if you dont do a null check here?

test/batchReceiver.spec.ts
```diff
@@ -234,31 +235,32 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
 
-  it("Abandoning a message 10 times, you get it no more. You can then only get it from the dead letter subscription", async function(): Promise<
+  it("Message abandoned more than maxDeliveryCount goes to dead letter subscriptions", async function(): Promise<
     void
   > {
     await topicClient.send(testMessages[0]);
     let abandonMsgCount = 0;
-    let receivedMsgs;
-    let clientSubscription: SubscriptionClient;
+    let deadletterSubscriptionClient: SubscriptionClient;
```
ramya-rao-a: Do we have to declare this here? Can we not declare it as a constant when we create the queue client?

test/batchReceiver.spec.ts
```diff
@@ -304,35 +303,32 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    const receivedMsgs = await queueClient.receiveBatch(1);
+    should.equal(Array.isArray(receivedMsgs), true);
+    should.equal(receivedMsgs.length, 1);
+    should.equal(receivedMsgs[0].body === testMessages[0].body, false);
+    should.equal(receivedMsgs[0].messageId === testMessages[0].messageId, false);
+    await receivedMsgs[0].complete();
+
+    const deferredMsgs = await queueClient.receiveDeferredMessage(sequenceNumber);
+    if (!deferredMsgs) {
+      throw "The message that we are receving from the sequence number can not be null";
```
ramya-rao-a: Can we simplify this? How about "No deferred message received for sequence number" ?

test/batchReceiver.spec.ts
ShivangiReja: Did you mean globally? outside of the method?

ramya-rao-a: No no :) Just move it to where it is actually being used

test/batchReceiver.spec.ts
```diff
@@ -236,7 +236,7 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
     await testPeekMsgsLength(queueClient, 0);
 
     const deadLetterQueuePath = Namespace.getDeadLetterQueuePathForQueue(
-      process.env.QUEUE_NAME ? process.env.QUEUE_NAME : ""
+      queueClient.name ? queueClient.name : ""
```
ramya-rao-a: The `name` property is not optional, so you can skip the null check here. Same for topic/subscription client

test/batchReceiver.spec.ts
```diff
@@ -409,13 +406,14 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
     await testPeekMsgsLength(subscriptionClient, 0);
 
     const deadLetterSubscriptionPath = Namespace.getDeadLetterSubcriptionPathForSubcription(
-      process.env.TOPIC_NAME ? process.env.TOPIC_NAME : "",
-      process.env.SUBSCRIPTION_NAME ? process.env.SUBSCRIPTION_NAME : ""
+      topicClient.name ? topicClient.name : "",
+      subscriptionClient.subscriptionName ? subscriptionClient.subscriptionName : ""
     );
 
+    let deadletterSubscriptionClient: SubscriptionClient;
```
ramya-rao-a: This can be a constant if you combine this line and the next one

Issue #145, opened 2019-01-04
Completing a message from a streaming receiver with autoComplete enabled results in program not exiting


ramya-rao-a:
**Describe the bug**
Completing a message from a streaming receiver with autoComplete enabled results in program not exiting

**To Reproduce**

- Use a streaming receiver (by default autoComplete will be enabled)
- Complete the message in the `onMessage` callback (`await message.complete()`)

This will result in the program not exiting. 

**Expected behavior**

Program exits 

**Additional context**

This is most likely due to the `autoComplete` feature trying to complete an already completed message

HarshaNalluru:
For me, it took more time to exit(about 10-15 seconds more) when added `await message.complete()` call. 
But I didn't have to manually exit in either case(with and without `await message.complete()`).
Did you have to manually kill/close the execution ?

HarshaNalluru:
**Bug**
   - It takes 20 seconds more time for the program to exit when we complete the message in the `onMessage` callback(`await message.complete()`) and if the autoComplete is true(by default autoComplete will be enabled).

**Reason for the Bug(Summary)**
   - When autoComplete is true(by default autoComplete will be enabled), `complete()` gets called for the second time and [timer attribute of _deliveryDispositionMap](https://github.com/Azure/azure-service-bus-node/blob/18f114dbfa0395b94de7380d8af9ceac26e8b560/lib/core/messageReceiver.ts#L781)(`settimeout()`) gets overwritten by another `settimeout()`. 
   - The initial `settimeout()` is not cleared but overwritten. This causes the program to wait for another 20 seconds(`messageDispositionTimeout = 20000`) and then exits.
---------------------------------
**Detailed Explanation**
- [complete()](https://github.com/Azure/azure-service-bus-node/blob/18f114dbfa0395b94de7380d8af9ceac26e8b560/lib/serviceBusMessage.ts#L851) method in the serviceBusMessage.ts gets called twice for each message. 
   - First time, we call it from the sample as `brokeredMessage.complete()`.
   - And if autoComplete is true(which is by default), [complete()](https://github.com/Azure/azure-service-bus-node/blob/18f114dbfa0395b94de7380d8af9ceac26e8b560/lib/core/messageReceiver.ts#L479
) is called for the second time from messageReceiver.ts.
[...]

PR #154, opened 2019-01-08
Fixing the bug - Completing an already completed message

HarshaNalluru:
## Description
**Issue** - https://github.com/Azure/azure-service-bus-node/issues/145

**Steps to reproduce the behavior**
- Use a streaming receiver (by default autoComplete will be enabled)
- Complete the message in the onMessage callback (`await message.complete()`)
- The program takes more time to exit(about 20 seconds more) when added `await message.complete()` call in the onMessage callback.

**Observations**
- The actual workflow with `await message.complete()` call in the onMessage callback is 
    - `complete() -> _onSettled() -> second complete()`.
- The reason for taking more time is [timer attribute of _deliveryDispositionMap(settimeout())]( https://github.com/Azure/azure-service-bus-node/blob/18f114dbfa0395b94de7380d8af9ceac26e8b560/lib/core/messageReceiver.ts#L781) is not getting cleared [Which actually happens in the ` _onSettled() ` event handler] 
   -  `_onSettled()` is not executed after the second `complete()` call[that came from the autoComplete being true by default].
- We don't have to call complete() method for the second time if the message gets settled by the first call.
- `delivery.remote_settled` gets updated to true when the message is settled.
-  So, we can check the property(`delivery.remote_settled`) just before making the second complete() call and make the call if that property is not true.
-------------------------------

**Changes in the PR**
- If `autoComplete` is `true`, `await bMessage.complete()` gets executed all the time without checking the  `delivery.remote_settled`. 
- We'd have to check the `delivery.remote_settled` and if it is false, then execute `await bMessage.complete()` 

The second `complete()` call.
https://github.com/Azure/azure-service-bus-node/blob/18f114dbfa0395b94de7380d8af9ceac26e8b560/lib/core/messageReceiver.ts#L479 
So, the above line of code changes to
```typescript 
if (!bMessage.delivery.remote_settled) {
       await bMessage.complete();
}
```
This would prevent calling `complete()` on a settled message.
- https://github.com/Azure/azure-service-bus-node/issues/145

@AlexGhiondea @ramya-rao-a @ShivangiReja @ramya0820

### DIFF
--- a/lib/core/messageReceiver.ts
+++ b/lib/core/messageReceiver.ts
@@ -468,7 +468,11 @@ export class MessageReceiver extends LinkEntity {
 
       // If we've made it this far, then user's message handler completed fine. Let us try
       // completing the message.
-      if (this.autoComplete && this.receiveMode === ReceiveMode.peekLock) {
+      if (
+        this.autoComplete &&
+        this.receiveMode === ReceiveMode.peekLock &&
+        !bMessage.delivery.remote_settled
+      ) {
         try {
           log[this.receiverType](
             "[%s] Auto completing the message with id '%s' on " + "the receiver '%s'.",
@@ -476,9 +480,7 @@ export class MessageReceiver extends LinkEntity {
             bMessage.messageId,
             this.name
           );
-          if (!bMessage.delivery.remote_settled) {
-            await bMessage.complete();
-          }
+          await bMessage.complete();
         } catch (completeError) {
           const translatedError = translate(completeError);
           log.error(
--- a/lib/core/messageReceiver.ts
+++ b/lib/core/messageReceiver.ts
@@ -476,7 +476,9 @@ export class MessageReceiver extends LinkEntity {
             bMessage.messageId,
             this.name
           );
-          await bMessage.complete();
+          if (!bMessage.delivery.remote_settled) {
+            await bMessage.complete();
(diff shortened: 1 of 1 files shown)


lib/core/messageReceiver.ts
```diff
@@ -476,7 +476,9 @@ export class MessageReceiver extends LinkEntity {
             bMessage.messageId,
             this.name
           );
-          await bMessage.complete();
+          if (!bMessage.delivery.remote_settled) {
```
ramya-rao-a: Shouldnt this check move to line 471? Otherwise, we will be logging that we are auto-completing when we dont want to auto-complete an already settled message

ramya-rao-a: This branch was out of date from master, I have updated it. So pull before making your changes

HarshaNalluru: This way ?
```typescript
if (this.autoComplete && this.receiveMode === ReceiveMode.peekLock && !bMessage.delivery.remote_settled)
```

Yeah, it makes sense. 
I'll just check to make sure I don't face any issues with that.

ramya-rao-a: Yes

ramya-rao-a:
Good point, yes.

Issue #101, opened 2018-12-19
Implement tests for Streaming Receiver in PeekLock mode


AlexGhiondea:
- [x] With auto-complete enabled + no message settling by the user should remove the message from the Queue/Subscription
- [x] With auto-complete disabled, manual completion by the user should remove the message from the Queue/Subscription
- [x] With auto-complete disabled, dead lettering the message results in not getting the same message again from queue/subscription. The message is then gotten only from the dead letter queue
- [x] With auto-complete disabled, abandoning a message results in getting the same message from queue/subscription with incremented delivery count. After 10 times, you get it no more. You can then only get it from the dead letter queue
- [x] With auto-complete enabled + manual 
   - [x] completion
   - [x] abandon
   - [x] defer 
   - [x] deadletter 
in Queue/Subscription by the user should not result in SDK trying to complete it again and no errors should be thrown
- [x] Second Streaming Receiver call should fail if the first one is not stopped
- [ ] Settling a message twice should throw error. (Blocked by #74)
- [x] With auto-complete disabled, deferring a message results in not getting the same message again from queue/subscription. The message is then gotten using `receiveDefferedMessages`
- [x] With auto-complete disabled, no settling of the message results in getting the same message again from queue/subscription
- [ ] Test `maxConcurrentCalls`. When set to 1 (default value), next message is received only after the first one is settled. Test with non default value as well. (Tracked in #179)

HarshaNalluru:
**All the tests here are done except** 
- Settling a message twice should throw error. (Blocked by #74)
------------------------------------
**Issues I've faced while working on tests:** 
- Receiving messages from a deadletter gives a timeout ( happens infrequently )
   - If working with a ServiceBus namespace gives the timeout errors, using a different ServiceBus (different namespace) helped getting rid of those timeouts.
- Clearing the messages using `receive()` did not work(second `receive()` didn't get invoked). So, cleared the messages using `receiveBatch()` instead in some of the tests.
- Some of the tests only work with Partitioned Queues and Partitioned Subscriptions.
---------------------------
**I've checked the list of tests for `ReceiveBatch`(#100). The following are the ones are not covered in this issue. Do we need tests similar to them in this issue ?** 
- Multiple sequential ReceiveBatch using Queue/Subscription, ensure result from one is not affected by the other
- Cases for which we are not yet sure what the result should be, but we should definitely cover: 
[...]

PR #156, opened 2019-01-09
Streaming Receiver Tests 

HarshaNalluru:
## Description
1. **With auto-complete disabled, Abandoned message is retained in the Queue/Subscription with incremented deliveryCount.**
2. **After 10 times, you can only get it from the dead letter.**
- `msg.abandon();` is called in the onMessage callback.
- This test checks
   - the incremented deliveryCount(verified inside the `onMessage` callback at every receiver instance)
   - when `deliveryCount` reaches `maxDeliveryCount(=10)`, the message is no longer present in the queue/subscription. Message can be retrieved only from the dead letter.
   - The `deliveryCount` attribute is also verified when the messages are received from dead letter queue. 

**Queue** https://github.com/Azure/azure-service-bus-node/blob/f7ae5234944e3f2536569f361d26fd61c3bb1683/test/streamingReceiver.spec.ts#L253
**Subscription** https://github.com/Azure/azure-service-bus-node/blob/f7ae5234944e3f2536569f361d26fd61c3bb1683/test/streamingReceiver.spec.ts#L310
- https://github.com/Azure/azure-service-bus-node/issues/101


HarshaNalluru:
**With auto-complete enabled, manual completion in the Queue/Subscription by the user should not result in errors**

**Queue**
https://github.com/Azure/azure-service-bus-node/blob/5d4714b9a89890a3fe14b00fa690c7e1d985ccf9/test/streamingReceiver.spec.ts#L369

**Subscription**
https://github.com/Azure/azure-service-bus-node/blob/5d4714b9a89890a3fe14b00fa690c7e1d985ccf9/test/streamingReceiver.spec.ts#L393

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,166 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+      (msg: ServiceBusMessage) => {
+        if (msg.messageId === testMessages[0].messageId) {
+          should.equal(msg.deliveryCount, checkDeliveryCount0);
+          checkDeliveryCount0++;
+        } else if (msg.messageId === testMessages[1].messageId) {
+          should.equal(msg.deliveryCount, checkDeliveryCount1);
+          checkDeliveryCount1++;
+        }
+        msg.abandon();
+        return Promise.resolve();
```
ramya-rao-a: Shouldnt we be doing `return msg.abandon()` instead?

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,166 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+    await delay(4000);
+
+    await receiveListener.stop();
+
+    should.equal(checkDeliveryCount0, maxDeliveryCount);
+    should.equal(checkDeliveryCount1, maxDeliveryCount);
+
+    peekedMsgs = await queueClient.peek(2);
+    should.equal(peekedMsgs.length, 0); // No messages in the queue
```
ramya-rao-a: You can replace the above 2 lines with `testPeekMsgsLength(queueClient, 0)`

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,166 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+  it("Abandoned message is retained in the Queue with incremented deliveryCount. After 10 times, you can only get it from the dead letter queue.", async function(): Promise<
+    void
+  > {
+    await queueClient.sendBatch(testMessages);
+
+    let peekedMsgs = await queueClient.peek(2);
+    should.equal(peekedMsgs.length, 2);
+    should.equal(peekedMsgs[0].deliveryCount, 0);
+    should.equal(peekedMsgs[1].deliveryCount, 0);
```
ramya-rao-a: Are the above checks needed? When you start receiving the messages you check for the deliverycount, so we can probably skip the peeking above

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,166 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    should.equal(checkDeliveryCount0, maxDeliveryCount);
+    should.equal(checkDeliveryCount1, maxDeliveryCount);
+
+    peekedMsgs = await queueClient.peek(2);
+    should.equal(peekedMsgs.length, 0); // No messages in the queue
+
+    const deadLetterQueuePath = Namespace.getDeadLetterQueuePathForQueue(queueClient.name);
+    const deadletterQueueClient = namespace.createQueueClient(deadLetterQueuePath);
+
+    await testPeekMsgsLength(deadletterQueueClient, 2); // Two messages in the DLQ
```
ramya-rao-a: Since you receiving 2 messages from the deadletter queue, this check can be skipped.

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,166 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+    await testPeekMsgsLength(deadletterSubscriptionClient, 0);
+  });
+
+  it("With auto-complete enabled, manual completion in the Queue by the user should not result in errors", async function(): Promise<
+    void
+  > {
+    await queueClient.sendBatch(testMessages);
+    let peekedMsgs = await queueClient.peek(2);
+    should.equal(peekedMsgs.length, 2); // Two messages in the queue
```
ramya-rao-a: Replace the above 2 lines with `testPeekMsgsLength(queueClient, 2)`

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,166 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+        should.not.exist(err);
+      }
+    );
+
+    await delay(4000);
+
+    await receiveListener.stop();
+
+    peekedMsgs = await queueClient.peek(2);
+    should.equal(peekedMsgs.length, 0); // No messages in the queue
```
ramya-rao-a: Replace the above 2 lines with `testPeekMsgsLength(queueClient, 0)`

HarshaNalluru:
**With auto-complete enabled, manual abandon in the Queue/Subscription by the user should not result in errors**

**Queue**
https://github.com/Azure/azure-service-bus-node/blob/ad5d1a637322121ff267ed38b5a68c713e8b5b32/test/streamingReceiver.spec.ts#L401

**Subscription**
https://github.com/Azure/azure-service-bus-node/blob/ad5d1a637322121ff267ed38b5a68c713e8b5b32/test/streamingReceiver.spec.ts#L429

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,220 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    await testPeekMsgsLength(queueClient, 0); // No messages in the queue
+
+    const deadLetterQueuePath = Namespace.getDeadLetterQueuePathForQueue(queueClient.name);
+    const deadletterQueueClient = namespace.createQueueClient(deadLetterQueuePath);
+
+    const deadLetterMsgs = await deadletterQueueClient.receiveBatch(2);
+    should.equal(Array.isArray(deadLetterMsgs), true);
+    should.equal(deadLetterMsgs.length, 2);
+    should.equal(deadLetterMsgs[0].deliveryCount, maxDeliveryCount);
+    should.equal(deadLetterMsgs[1].deliveryCount, maxDeliveryCount);
```
ramya-rao-a: Here lets also test the message id. Otherwise this test will pass if there were other messages in the dead letter queue with same deliver count

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,220 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+    const deadLetterSubscriptionPath = Namespace.getDeadLetterSubcriptionPathForSubcription(
+      topicClient.name,
+      subscriptionClient.subscriptionName
+    );
+
+    const deadletterSubscriptionClient = namespace.createSubscriptionClient(
+      deadLetterSubscriptionPath ? deadLetterSubscriptionPath : "",
+      subscriptionClient.subscriptionName
+    );
```
ramya-rao-a: Shouldn't the first parameter here be the topic path?

test/streamingReceiver.spec.ts
```diff
@@ -247,4 +249,220 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    await testPeekMsgsLength(subscriptionClient, 0);
+  });
+
+  it("With auto-complete enabled, manual abandon in the Queue by the user should not result in errors", async function(): Promise<
+    void
+  > {
+    await queueClient.sendBatch(testMessages);
+    const receiveListener = await queueClient.receive(
+      (msg: ServiceBusMessage) => {
+        return msg.abandon();
```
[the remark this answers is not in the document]
HarshaNalluru: How can I stop the receiveListener right after the first abandon() call ?

test/streamingReceiver.spec.ts
HarshaNalluru: `deadLetterSubscriptionPath` gives the path to the DeadLetter which has topic in it.(`<topic>/Subscriptions/<subscription>/$DeadLetterQueue`).

Giving the `topicClient` name is no different than the usual way of receiving the messages.

test/streamingReceiver.spec.ts
HarshaNalluru: Made commits for this. Working fine.
A small question : 
Is it possible that the two messages in the deadletter are in different order compared to the sent messages ?

test/streamingReceiver.spec.ts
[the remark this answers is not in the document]
ramya-rao-a: same way as you do on the outside :)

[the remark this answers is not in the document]
HarshaNalluru: Removed this test for both Queues and Subscriptions for now.
Verified all the previous four tests after the changes and made the commit.

HarshaNalluru:
Verified all the previous four tests after the changes and then made the commit.
@ramya-rao-a 
Please merge this to the master if you don't find any other issues with this.

PR #158, opened 2019-01-10
Batch Receiver Tests

ShivangiReja:
## Description

Implemented following tests for ReceiveBatch : 

 1. No settling of the message results in getting the same message again (with delivery count incremented) from queue/subscription after the lock duration is expired.
 2.   Ask for `n` msgs, but queue/subscription only has `m` msgs where m < n. This should return exactly `m` messages.
 3.  Abandon/Deadletter/Defer a message received from recevieDefferedMessage using queue/subscription: 
      -  Abandoning a message received from recevieDefferedMessage results in getting the same 
         message again from queue/subscription using `receiveDefferedMessages`.
     -  Deferring a message  received from recevieDefferedMessage results in not getting the same 
        message again from queue/subscription. The message is then gotten using 
        `receiveDefferedMessages`
     - Dead lettering a message received from recevieDefferedMessage results in not getting the same 
        message again from queue/subscription using recevieDefferedMessage  . The message is then 
        gotten only from the dead letter queue.
 4.  Abandon/Deadletter/Defer a message received from the dead letter queue/ subscription :
     -  Abandoning a message received from the dead letter queue/ subscriptionresults in getting the 
       same message again from the dead letter queue/subscription.
     -  Deferring a message  received  from the dead letter queue/ subscription results in not getting the 
        same message again from dead queue/subscription. The message is then gotten using 
        `receiveDefferedMessages`
     - Dead lettering a message received from the dead letter queue/ subscription should results error.

### DIFF
--- a/test/batchReceiver.spec.ts
+++ b/test/batchReceiver.spec.ts
@@ -43,6 +43,7 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
   let queueClient: QueueClient;
   let topicClient: TopicClient;
   let subscriptionClient: SubscriptionClient;
+  let errorWasThrown: boolean;
 
   beforeEach(async () => {
     // The tests in this file expect the env variables to contain the connection string and
@@ -82,6 +83,7 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
     if (peekedSubscriptionMsg.length) {
       throw new Error("Please use an empty Subscription for integration testing");
     }
+    errorWasThrown = false;
   });
 
   afterEach(async () => {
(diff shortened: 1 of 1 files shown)


test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+  });
+
+  it("No settlement of the message is retained in the Subscriptions with incremented deliveryCount", async function(): Promise<
+    void
+  > {
+    await topicClient.send(testMessages[0]);
+    let receivedMsgs = await subscriptionClient.receiveBatch(1);
+
+    should.equal(receivedMsgs.length, 1);
+    should.equal(receivedMsgs[0].body, testMessages[0].body);
```
ramya-rao-a: Shouldn't we be testing the delivery count to be 0 here?

test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    should.equal(receivedMsgs[0].messageId, testMessages[0].messageId);
+
+    await testPeekMsgsLength(subscriptionClient, 1);
+
+    receivedMsgs = await subscriptionClient.receiveBatch(1);
+
+    should.equal(receivedMsgs.length, 1);
+    should.equal(receivedMsgs[0].body, testMessages[0].body);
```
ramya-rao-a: Shouldn't we be testing the delivery count to be 1 here?

test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+      throw "No message received for sequence number";
+    }
+    should.equal(deferredMsgs.body, testMessages[0].body);
+    should.equal(deferredMsgs.messageId, testMessages[0].messageId);
+
+    await deferredMsgs.abandon();
+
+    await testPeekMsgsLength(queueClient, 1);
+
+    if (!sequenceNumber) {
```
ramya-rao-a: This check is already made in `deferMsgAndGetSequenceNum`. Do we need to do it again here?

test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+    if (!receivedMsgs[0].sequenceNumber) {
+      throw "Sequence Number can not be null";
+    }
+    const sequenceNumber = receivedMsgs[0].sequenceNumber;
+    await receivedMsgs[0].defer();
+    return sequenceNumber;
+  }
+
+  it("Abandon a message received from recevieDefferedMessage using queue", async function(): Promise<
```
ramya-rao-a: Change this to `Abandoning a deferred message returns it to deferred queue`
Same for subscription

test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    }
+    should.equal(abandonedMsgs.body, testMessages[0].body);
+    should.equal(abandonedMsgs.messageId, testMessages[0].messageId);
+
+    await abandonedMsgs.complete();
+
+    await testPeekMsgsLength(subscriptionClient, 0);
+  });
+
+  it("Dead letter a message received from recevieDefferedMessage using queue", async function(): Promise<
```
ramya-rao-a: Change this to `Deadlettering a deferred message moves it to dead letter queue`.
Same for subscription

test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+    const deadLetterSubscriptionPath = Namespace.getDeadLetterSubcriptionPathForSubcription(
+      topicClient.name,
+      subscriptionClient.subscriptionName
+    );
+
+    const deadletterSubscriptionClient = namespace.createSubscriptionClient(
+      deadLetterSubscriptionPath ? deadLetterSubscriptionPath : "",
+      subscriptionClient.subscriptionName
+    );
```
ramya-rao-a: We seem to be repeating this a lot :) First calling `Namespace.getDeadLetterSubcriptionPathForSubcription` then `namespace.createSubscriptionClient`. Maybe we can have a helper method in the test file that combines both?

test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+    if (!sequenceNumber) {
+      throw "Sequence Number can not be null";
+    }
+    const abandonedMsgs = await queueClient.receiveDeferredMessage(sequenceNumber);
+    if (!abandonedMsgs) {
+      throw "No message received for sequence number";
+    }
+    should.equal(abandonedMsgs.body, testMessages[0].body);
+    should.equal(abandonedMsgs.messageId, testMessages[0].messageId);
```
ramya-rao-a: Since this was an abandoned message, can we check that the delivery count got incremented?

test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+
+    should.equal(deadLetterMsgs.length, 1);
+    should.equal(deadLetterMsgs[0].body, testMessages[0].body);
+    should.equal(deadLetterMsgs[0].messageId, testMessages[0].messageId);
+
+    await deadLetterMsgs[0].complete();
+    await testPeekMsgsLength(deadletterSubscriptionClient, 0);
+  });
+
+  it("Defer a message received from recevieDefferedMessage using queue", async function(): Promise<
```
ramya-rao-a: Rename to `Deferring a deferred message puts it back to the deferred queue`.

The delivery count should not be incremented in this case, correct?

test/batchReceiver.spec.ts
```diff
@@ -421,4 +423,533 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
...
+    should.equal(deadLetterMsgs[0].messageId, testMessages[0].messageId);
+
+    await deadLetterMsgs[0].abandon();
+
+    const abandonedMsgs = await deadletterQueueClient.receiveBatch(1);
+
+    should.equal(abandonedMsgs.length, 1);
+    should.equal(abandonedMsgs[0].body, testMessages[0].body);
+    should.equal(abandonedMsgs[0].messageId, testMessages[0].messageId);
+
```
ramya-rao-a: Since we are abandoning the message, we should check the delivery count

test/batchReceiver.spec.ts
ShivangiReja: No, I'll remove it from here.

test/batchReceiver.spec.ts
ShivangiReja: I'll make 2 helper methods one for queue and one for subscription.

test/batchReceiver.spec.ts
ShivangiReja: yes, I'll add one check for delivery count.

test/batchReceiver.spec.ts
ShivangiReja: Yes, It increments the delivery count. I'll add one check for delivery count here.

test/batchReceiver.spec.ts
ShivangiReja: I don't know why but abandoning the message received from dead letter queue does not increase the delivery count.

ramya-rao-a: That is very interesting, maybe that is intentional. Check with Neeraj if that is expected. Just send a message in the Service Bus channel in Teams

PR #164, opened 2019-01-11
Streaming Receiver Tests - Part 3

HarshaNalluru:
**With auto-complete enabled, manual abandon/defer in the Queue/Subscription by the user should not result in errors.** (Using Partitioned Queues and Subscriptions)
- abandon(Queue)
- abandon(Subscription)
- defer(Queue)
- defer(Subscription)

**Note :**  Clearing the messages using `receive()` did not work(second receive() didn't get invoked). So, cleared the messages using `receiveBatch()` instead.

- #101


HarshaNalluru:
Added tests for
`Second Streaming Receiver call should fail if the first one is not stopped for Queues/Subscriptions`

PR #171, opened 2019-01-14
Refactoring StreamingReceiver Tests to re-use code across test cases

ramya-rao-a:
### Description

Net changes in this PR:
- FIFO is not guaranteed unless we use sessions. Tests that expect `receivedMsgs` and `testMessages` to have contents in the same order are updated accordingly
- Tests are broken into separate logical chunks. This will help future maintaince
- Re-use test code between Queues and Subscriptions, autocomplete enable/disable
- Repeated code is moved to a common place
- Test descriptions updated to match the ones in the tests for the batching receiver
- Increased the time that we wait before stopping the receiveListener as it kept getting stopped before the messages were received.
- Add tests for abandon() when autoComplete is disabled

### DIFF
--- a/test/streamingReceiver.spec.ts
+++ b/test/streamingReceiver.spec.ts
@@ -587,7 +587,7 @@ describe("With autocomplete enabled, test Complete/Abandon/Defer/Deadletter norm
   });
 });
 
-describe.only("With autocomplete disabled, test Complete/Abandon/Defer/Deadletter normal message", function(): void {
+describe("With autocomplete disabled, test Complete/Abandon/Defer/Deadletter normal message", function(): void {
   beforeEach(async () => {
     await beforeEachTest();
   });
--- a/test/streamingReceiver.spec.ts
+++ b/test/streamingReceiver.spec.ts
@@ -324,251 +324,7 @@ describe("Streaming Receiver Misc Tests", function(): void {
   });
 });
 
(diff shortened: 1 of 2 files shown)


test/streamingReceiver.spec.ts
```diff
@@ -657,137 +496,109 @@ describe("Streaming Receiver from Queue/Subscription", function(): void {
...
-    await testPeekMsgsLength(deadletterSubscriptionClient, 0);
+  afterEach(async () => {
+    await afterEachTest();
   });
 
-  it("Second Streaming Receiver call should fail if the first one is not stopped for Queues", async function(): Promise<
-    void
-  > {
-    const receiveListener: ReceiveHandler = await queueClient.receive(
+  async function testMultipleReceiveBatchCalls(
```
HarshaNalluru: The name `testMultipleReceiveBatchCalls` sounds like it belongs to the batch receiver.
Should we change that to `testMultipleReceiveCalls` or something ?

HarshaNalluru:
I've executed the tests and verified everything works.
I don't see any issues with this.
Merge this PR before you merge https://github.com/Azure/azure-service-bus-node/pull/170.
Once this gets merged, I'll resolve the merge conflicts for the https://github.com/Azure/azure-service-bus-node/pull/170.

PR #174, opened 2019-01-14
ReceiveAndDelete mode Tests

ShivangiReja:
Implement following test cases:

- ReceiveBatch from Queue/Subscription with no settling from the user should remove the message from the queue
-  Streaming Receiver with no settling from the user should remove the message from the queue/subscription regardless of whether autoComplete is enabled or not
- complete(), deadletter(), abandon()- Before calling these methods, messages already removed from the queue/subscription using receive method, so it does nothing.
- defer(), renewLock() - Throws error.

### DIFF
--- a/test/receiveAndDeleteMode.spec.ts
+++ b/test/receiveAndDeleteMode.spec.ts
@@ -1,0 +1,368 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import "mocha";
+import * as chai from "chai";
+const should = chai.should();
+import * as chaiAsPromised from "chai-as-promised";
+import * as dotenv from "dotenv";
+dotenv.config();
+chai.use(chaiAsPromised);
+import {
+  Namespace,
+  QueueClient,
+  SendableMessageInfo,
+  generateUuid,
+  TopicClient,
+  SubscriptionClient,
+  delay,
+  ServiceBusMessage,
+  ReceiveMode
+} from "../lib";
+
+const testMessages: SendableMessageInfo[] = [
+  {
+    body: "hello1",
+    messageId: `test message ${generateUuid()}`
+  },
+  {
+    body: "hello2",
+    messageId: `test message ${generateUuid()}`
+  }
+];
+
+async function testPeekMsgsLength(
+  client: QueueClient | SubscriptionClient,
+  expectedPeekLength: number
+): Promise<void> {
+  const peekedMsgs = await client.peek(expectedPeekLength + 1);
+  should.equal(peekedMsgs.length, expectedPeekLength);
+}
+
+let namespace: Namespace;
+let queueClient: QueueClient;
+let topicClient: TopicClient;
+let subscriptionClient: SubscriptionClient;
+let deadletterQueueClient: QueueClient;
+let deadletterSubscriptionClient: SubscriptionClient;
+let errorWasThrown: boolean;
+
+async function beforeEachTest(): Promise<void> {
+  // The tests in this file expect the env variables to contain the connection string and
+  // the names of empty queue/topic/subscription that are to be tested
+
+  if (!process.env.SERVICEBUS_CONNECTION_STRING) {
+    throw new Error(
+      "Define SERVICEBUS_CONNECTION_STRING in your environment before running integration tests."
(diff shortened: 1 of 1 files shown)


test/receiveAndDeleteMode.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+    messageId: `test message ${generateUuid()}`
+  }
+];
+
+async function testPeekMsgsLength(
+  client: QueueClient | SubscriptionClient,
+  expectedPeekLength: number
+): Promise<void> {
+  const peekedMsgs = await client.peek(expectedPeekLength + 1);
+  should.equal(peekedMsgs.length, expectedPeekLength);
```
ramya-rao-a: Please add a message here like we do in https://github.com/Azure/azure-service-bus-node/blob/master/test/batchReceiver.spec.ts#L41
This will help us in figuring out which check failed

test/receiveAndDeleteMode.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+    process.env.SUBSCRIPTION_NAME,
+    {
+      receiveMode: ReceiveMode.receiveAndDelete
+    }
+  );
+  deadletterQueueClient = namespace.createQueueClient(
+    Namespace.getDeadLetterQueuePathForQueue(queueClient.name)
+  );
+  deadletterSubscriptionClient = namespace.createSubscriptionClient(
+    Namespace.getDeadLetterSubcriptionPathForSubcription(
```
ramya-rao-a: Dont we need the `deadletterQueueClient` and the `deadletterSubscriptionClient` to be in ReceiveAndDelete mode as well?

test/receiveAndDeleteMode.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+  it("Queue: No settlement of the message removes message", async function(): Promise<void> {
+    await testNosettlment(queueClient, queueClient);
+  });
+
+  it("Subscription: No settlement of the message removes message", async function(): Promise<void> {
+    await testNosettlment(topicClient, subscriptionClient);
+  });
+});
+
+describe("Streaming Receiver from Queue/Subscription", function(): void {
```
ramya-rao-a: The tests inside this block needs to be repeated with autoComplete disabled.
Pass `autoComplete` as a variable to `sendReceiveMsg`, then you can re-use that code

test/receiveAndDeleteMode.spec.ts
ShivangiReja: Yes, I've used deadletterQueueClient and the deadletterSubscriptionClient for deadletter test case.

test/receiveAndDeleteMode.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+    should.equal(Array.isArray(msgs), true);
+    should.equal(msgs.length, 1);
+    should.equal(msgs[0].body, testMessages[0].body);
+    should.equal(msgs[0].messageId, testMessages[0].messageId);
+    should.equal(msgs[0].deliveryCount, 0);
+
+    return msgs[0];
+  }
+
+  async function testComplete(
```
ramya-rao-a: There isnt much difference between `testComplete()`, `testAbandon()` etc.

Why not rename this to `testSettlement`, make it take another parameter `dispositionType` which would be an enum [DispositionType](https://github.com/Azure/azure-service-bus-node/blob/981c8472b6d94da47607123f3023be5da322e91a/lib/core/messageReceiver.ts#L68). Then you can have switch statement inside to do either of complete()/abandon()/.....

test/receiveAndDeleteMode.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+    await testDeadletter(topicClient, subscriptionClient);
+  });
+
+  async function sendReceiveMsgWithRenewLock(
+    senderClient: QueueClient | TopicClient,
+    receiverClient: QueueClient | SubscriptionClient
+  ): Promise<void> {
+    await senderClient.sendBatch(testMessages);
+    const receivedMsgs: ServiceBusMessage[] = [];
+    const receiveListener = receiverClient.receive(
```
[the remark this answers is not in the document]
ShivangiReja: No, I'll change it.

Issue #103, opened 2018-12-19
Implement test for Lock Renewal in PeekLock  mode


AlexGhiondea:
- [x] Receive a msg using ReceiveBatch, wait until its lock expires, completing it now results in error
- [x] Receive a msg using ReceiveBatch, test renewLock()
- [x] Receive a msg using Streaming, disable auto lock renewal, use .renewLock() manually (set maxLockDuration to 0)
- [x] Receive a msg using Streaming, Test auto lock renewal (include the time until which auto lock renewal is suppose to run)
- [x] Auto lock renewal: Dont renew if msg lock expiry time is longer than the configured maxLockDuration

ramya-rao-a:
Done with https://github.com/Azure/azure-sdk-for-js/pull/970

PR #177, opened 2019-01-15
Added tests for Lock Renewal usecases #103

ramya0820:
## Description

- Added test cases based on discussion around #103

- #103


test/renewLock.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  should.equal(Array.isArray(msgs), true);
+  should.equal(msgs.length, 1);
+  should.equal(msgs[0].body, testMessages[0].body);
+  should.equal(msgs[0].messageId, testMessages[0].messageId);
+
+  console.log(`MessageLockedUntil: ${msgs[0].lockedUntilUtc}`);
+
+  // Compute expected initial lock duration
+  currentExpectedLockExpiryTimeUtc.setSeconds(
```
ramya-rao-a: Can we move this to right after `currentExpectedLockExpiryTimeUtc` is declared?
I was initially wondering that the current expected lock expiry time could not be the current time i.e.`new Date()` until I reached this line

test/renewLock.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+  });
+
+  it("TopicsSubs - Receive a msg using Batch Receiver, wait until its lock expires, completing it now results in error", async function(): Promise<
+    void
+  > {
+    await testErrorOnLockExpiry(topicClient, subscriptionClient);
+  });
+});
+
+async function testLockRenewalHappyCase(
```
ramya-rao-a: Can we move this just before we call it for the first time?
That's the pattern used for the other tests, and consistency helps for future maintenance and readability

test/renewLock.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+  await senderClient.sendBatch(testMessages);
+
+  const msgs = await receiverClient.receiveBatch(1);
+
+  should.equal(Array.isArray(msgs), true);
+  should.equal(msgs.length, 1);
+  should.equal(msgs[0].body, testMessages[0].body);
+  should.equal(msgs[0].messageId, testMessages[0].messageId);
+
+  console.log(`MessageLockedUntil: ${msgs[0].lockedUntilUtc}`);
```
ramya-rao-a: Console logs are generally not used in tests. Is there any reason you want to keep them?

test/renewLock.spec.ts
ramya0820: Agree, updated the variable name too.

Issue #153, opened 2019-01-08
Impement tests for add/remove/get Topic Filters


ramya-rao-a:
**Topic Filters: Get Rules**

- [x] Subscription with 0/1/multiple rules returns rules as expected
- [x] Default rule is returned for the subscription for which no rules were added
- [x] Rule with SQL filter returns expected filter expression
- [ ] Rule with SQL filter and action returns expected filter and action expression
- [x] Rule with Correlation filter returns expected filter
- [ ] Rule with Correlation filter and action (Is this allowed?) returns expected filter and action expression

**Topic Filters: Add Rule**

Positive Test Cases 
- [ ] Add Rule with Boolean filter 
- [x] Add Rule with SQL filter 
- [ ] Add Rule with SQL filter and action 
- [x] Add Rule with Correlation filter 
- [ ] Add Rule with Correlation filter and action (Is this allowed?) 

  
Negative Test Cases 
- [x] Adding a rule with a name which matches with existing rule 
- [x] Adding a rule with no name  
- [x] Adding a rule with no filter 
- [ ] Adding a rule with a filter that is neither boolean, string or a co-relation object
- [ ] Adding a rule with a SQL action whose input is not a string

**Topic Filters: Remove Rule**
- [x] Removing non existing rule on a subscription that doesnt have any rules should throw error
- [x] Removing non existing rule on a subscription that has other rules should throw error
- [x] Removing existing rule on a subscription

HarshaNalluru:
Whenever the rules are added or we see the existing rules, do we have to verify if the rule is working by sending and receiving the data ?
@ramya-rao-a

ramya-rao-a:
I wrote these test cases only for adding/removing/getting of rules, so we dont need to test if the rule actually works

But you are right we should have something to test if the rule actually works :)

Lets use this issue to track just the get/add/remove.
I'll log a separate issue for testing the rules themselves

ramya-rao-a:
The leftover test cases are being blocked by bugs #186 & #187

PR #181, opened 2019-01-15
TopicFilters - Tests

HarshaNalluru:
## Description
`sqlRuleActionExpression` is not working as per the expectations.
Verified all the tests which don't involve actions and committed the same.

- #153

### DIFF
--- a/test/topicFilters.spec.ts
+++ b/test/topicFilters.spec.ts
@@ -62,7 +62,7 @@ async function afterEachTest(): Promise<void> {
   await namespace.close();
 }
 
-describe("Topic Filters - Tests", function(): void {
+describe("Topic Filters -  Add Rule - Positive Test Cases", function(): void {
   beforeEach(async () => {
     await beforeEachTest();
   });
@@ -107,6 +107,16 @@ describe("Topic Filters - Tests", function(): void {
     const rules = await subscriptionClient.getRules();
     should.equal(rules.length, 1);
     should.equal(rules[0].name, "Correlationfilter");
+  });
+});
+
+describe("Topic Filters -  Add Rule - Negative Test Cases", function(): void {
+  beforeEach(async () => {
+    await beforeEachTest();
+  });
+
+  afterEach(async () => {
+    await afterEachTest();
   });
 
   it("Adding a rule with a name which matches with existing rule", async function(): Promise<void> {
@@ -159,6 +169,16 @@ describe("Topic Filters - Tests", function(): void {
     }
     should.equal(errorWasThrown, true);
   });
+});
+
+describe("Topic Filters -  Remove Rule", function(): void {
+  beforeEach(async () => {
+    await beforeEachTest();
+  });
+
+  afterEach(async () => {
+    await afterEachTest();
+  });
 
   it("Removing non existing rule on a subscription that doesnt have any rules should throw error", async function(): Promise<
     void
@@ -185,6 +205,16 @@ describe("Topic Filters - Tests", function(): void {
       errorWasThrown = true;
     }
     should.equal(errorWasThrown, true);
+  });
+});
+
+describe("Topic Filters: Get Rules", function(): void {
+  beforeEach(async () => {
+    await beforeEachTest();
+  });
+
+  afterEach(async () => {
+    await afterEachTest();
   });
 
   it("Subscription with 0/1/multiple rules returns rules as expected", async function(): Promise<
--- a/test/topicFilters.spec.ts
+++ b/test/topicFilters.spec.ts
@@ -12,14 +12,10 @@ import { Namespace, SubscriptionClient } from "../lib";
 
 // We need to remove rules before adding one because otherwise the existing default rule will let in all messages.
 async function removeAllRules(client: SubscriptionClient): Promise<void> {
-  try {
-    const rules = await client.getRules();
-    for (let i = 0; i < rules.length; i++) {
-      const rule = rules[i];
-      await client.removeRule(rule.name);
-    }
-  } catch (err) {
-    console.log("Error while removing the rule", err);
+  const rules = await client.getRules();
+  for (let i = 0; i < rules.length; i++) {
+    const rule = rules[i];
+    await client.removeRule(rule.name);
   }
 }
 
--- a/test/topicFilters.spec.ts
+++ b/test/topicFilters.spec.ts
@@ -99,17 +99,6 @@ describe("Topic Filters - Tests", function(): void {
     should.equal(rules[0].name, "Priority_1");
   });
 
-  it("Add Rule with SQL filter and action", async function(): Promise<void> {
-    await subscriptionClient.addRule(
-      "Priority_1",
-      "(priority = 1 OR priority = 3) AND (sys.label LIKE '%String1')",
-      "SET sys.body = 'MessageX'"
-    );
-    const rules = await subscriptionClient.getRules();
-    should.equal(rules.length, 1);
-    should.equal(rules[0].name, "Priority_1");
-  });
-
   it("Add Rule with Correlation filter", async function(): Promise<void> {
     await subscriptionClient.addRule("Correlationfilter", {
       label: "red",
@@ -118,18 +107,6 @@ describe("Topic Filters - Tests", function(): void {
     const rules = await subscriptionClient.getRules();
     should.equal(rules.length, 1);
     should.equal(rules[0].name, "Correlationfilter");
-  });
-
-  it("Add Default rule on a subscription that its default rule removed", async function(): Promise<
-    void
-  > {
-    await subscriptionClient.addRule("$Default", {
-      label: "red",
-      correlationId: "high"
-    });
-    const rules = await subscriptionClient.getRules();
-    should.equal(rules.length, 1);
-    should.equal(rules[0].name, "$Default");
   });
 
   it("Adding a rule with a name which matches with existing rule", async function(): Promise<void> {
--- a/test/topicFilters.spec.ts
+++ b/test/topicFilters.spec.ts
@@ -156,12 +156,12 @@ describe("Topic Filters -  Add Rule - Negative Test Cases", function(): void {
     should.equal(errorWasThrown, true);
   });
 
+  // TODO: Update error message after fixing https://github.com/Azure/azure-service-bus-node/issues/184
   it("Adding a rule with a Boolean filter whose input is not a Boolean, SQL expression or a Correlation filter", async function(): Promise<
     void
   > {
     let errorWasThrown = false;
     try {
-      await subscriptionClient.addRule("Priority_1", "priority = 1");
       await subscriptionClient.addRule("Priority_2", "1");
     } catch (error) {
       errorWasThrown = true;
--- a/test/topicFilters.spec.ts
+++ b/test/topicFilters.spec.ts
@@ -223,40 +223,31 @@ describe("Topic Filters: Get Rules", function(): void {
     let rules = await subscriptionClient.getRules();
     should.equal(rules.length, 0);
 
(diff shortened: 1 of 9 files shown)


test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+  });
+
+  afterEach(async () => {
+    await afterEachTest();
+  });
+
+  it("Add Rule with Boolean filter", async function(): Promise<void> {
+    await subscriptionClient.addRule("TrueFilter", true);
+    const rules = await subscriptionClient.getRules();
+    should.equal(rules[0].name, "TrueFilter");
```
ramya-rao-a: In every test where a rule is being added, check for the length of the `rules` array

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+    await subscriptionClient.addRule(
+      "Priority_1",
+      "(priority = 1 OR priority = 3) AND (sys.label LIKE '%String1')",
+      "SET sys.body = 'MessageX'"
+    );
+    const rules = await subscriptionClient.getRules();
+    should.equal(rules[0].name, "Priority_1");
+  });
+
+  it("Add Rule with SQL filter and action", async function(): Promise<void> {
```
ramya-rao-a: This test case is same as the one before it

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+  it("Add Rule with Correlation filter", async function(): Promise<void> {
+    await subscriptionClient.addRule("Correlationfilter", {
+      label: "red",
+      correlationId: "high"
+    });
+    const rules = await subscriptionClient.getRules();
+    should.equal(rules[0].name, "Correlationfilter");
+  });
+
+  it("Add Default rule on a subscription that its default rule removed", async function(): Promise<
```
ramya-rao-a: On second thoughts, this test case doesnt make sense. You can remove it.

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  it("Adding a rule with a name which matches with existing rule", async function(): Promise<void> {
+    await subscriptionClient.addRule("Priority_1", "priority = 1");
+    let errorWasThrown = false;
+    try {
+      await subscriptionClient.addRule("Priority_1", "priority = 2");
+    } catch (error) {
+      errorWasThrown = true;
+      should.equal(!error.message.search(" already exists."), false);
+      console.log(error.message);
```
ramya-rao-a: - Doesnt the error message contain the rule name? If so, then can we include that in the search here?
- Does the error have a name? If yes, then lets check for that as well.
- Why do we want to console.log the error?

Same applies to all other test cases where we are expecting an error

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  it("Adding a rule with a Boolean filter whose input is not Boolean", async function(): Promise<
+    void
+  > {
+    let errorWasThrown = false;
+    try {
+      await subscriptionClient.addRule("Priority_1", "priority = 1");
+      await subscriptionClient.addRule("Priority_2", "1");
+    } catch (error) {
+      errorWasThrown = true;
```
ramya-rao-a: What error is thrown here?

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+      should.equal(!error.message.search("Filter is missing"), false);
+      console.log(error.message);
+    }
+    should.equal(errorWasThrown, true);
+  });
+
+  it("Adding a rule with a Boolean filter whose input is not Boolean", async function(): Promise<
+    void
```
ramya-rao-a: Since we dont have separate function to add a "boolean" filter, this (and the commented test case after this) doesnt make sense....

How about changing this to `Adding a rule with a filter that is neither boolean, string or a co-relation object`

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+    process.env.DEFAULT_SUBSCRIPTION_NAME
+  );
+  await removeAllRules(subscriptionClient);
+}
+
+async function afterEachTest(): Promise<void> {
+  await namespace.close();
+}
+
+describe("Topic Filters - Tests", function(): void {
```
ramya-rao-a: Lets split the tests into different `describe` sections here. You can use the grouping in https://github.com/Azure/azure-service-bus-node/issues/153

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+    should.equal(JSON.stringify(rules[0].filter), JSON.stringify({ expression: expr }));
+
+    expr = "(priority = 1 OR priority = 3) AND (sys.label LIKE '%String1')";
+    await subscriptionClient.addRule(
+      "Priority_2",
+      "(priority = 1 OR priority = 3) AND (sys.label LIKE '%String1')"
+    );
+    rules = await subscriptionClient.getRules();
+    should.equal(rules[1].name, "Priority_2");
+    should.equal(JSON.stringify(rules[0].filter), JSON.stringify({ expression: expr }));
```
ramya-rao-a: The above 2 checks should be done for `rules[0]` as well

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+    should.equal(rules.length, 0);
+
+    let expr = "(priority = 1)";
+    await subscriptionClient.addRule("Priority_1", expr);
+    rules = await subscriptionClient.getRules();
+    should.equal(rules[0].name, "Priority_1");
+    should.equal(JSON.stringify(rules[0].filter), JSON.stringify({ expression: expr }));
+
+    expr = "(priority = 1 OR priority = 3) AND (sys.label LIKE '%String1')";
+    await subscriptionClient.addRule(
```
ramya-rao-a: Lets make the second rule have an action as well, and then add a check to see if the action actually got added to the rule

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+    rules = await subscriptionClient.getRules();
+    should.equal(rules[1].name, "Priority_2");
+    should.equal(JSON.stringify(rules[0].filter), JSON.stringify({ expression: expr }));
+  });
+
+  it("Default rule is returned for the subscription for which no rules were added", async function(): Promise<
+    void
+  > {
+    const rules = await defaultSubscriptionClient.getRules();
+    should.equal(rules[0].name, "$Default");
```
ramya-rao-a: Check length of `rules`

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+  });
+
+  it("Rule with SQL filter returns expected filter expression", async function(): Promise<void> {
```
ramya-rao-a: I wonder if we can skip this, as it is already covered in the test case for `Subscription with 0/1/multiple rules returns rules as expected`

test/topicFilters.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+    const expr = "(priority = 1 OR priority = 3) AND (sys.label LIKE '%String1')";
+    await subscriptionClient.addRule("Priority_2", expr);
+    const rules = await subscriptionClient.getRules();
+    should.equal(rules[0].name, "Priority_2");
+    should.equal(JSON.stringify(rules[0].filter), JSON.stringify({ expression: expr }));
+  });
+  /*
+  it("Rule with SQL filter and action returns expected filter and action expression", async function(): Promise<
```
ramya-rao-a: What happened here?

HarshaNalluru:
Comments on the Topic Filters Tests.
## Topic Filters 
**Positive Cases**
(Add Rule with Boolean filter)
- It doesn't let me add false
- https://docs.microsoft.com/en-us/azure/service-bus-messaging/topic-filters
- In the managementClient.ts, there's a check in the line 1211 addrule module, If(!filter){} -> "Cannot add rule. Filter is missing." When `false` is passed as boolean filter, it goes directly into the if block.
 
(Add Rule with SQL filter and action)
- Unable to change any property
- SQL Action thing is not working
 
 
(Add Rule with Correlation filter and action)
- SQL Action thing is not working, Unable to change any property

Adding a rule with a SQL filter whose input is not a string or is an empty string
- Comes under Adding a rule with no filter 
  
Adding a rule with a SQL action whose input is not a string
- Doesn't show error because the action is not working 

**Negative Cases**
Removing existing rule on a subscription
- Why should this give an error?

test/topicFilters.spec.ts
HarshaNalluru: Oops! I was testing the action part in this temporary code and forgot to remove this.

test/topicFilters.spec.ts
HarshaNalluru: For the test `"Adding a rule with a Boolean filter whose input is not Boolean"`

> { InternalServerError: The service was unable to process the request; please retry the operation. For more information on exception types and proper exception handling, please refer to http://go.microsoft.com/fwlink/?LinkId=761101, SystemTracker:NoSystemTracker, Timestamp:2019-01-15T19:03:45

Checked many times, this is the error I see. 
Are we expecting something like "Cannot accept this type of expression" ?

test/topicFilters.spec.ts
HarshaNalluru: For the test `"Adding a rule with a Boolean filter whose input is not Boolean"`

> { InternalServerError: The service was unable to process the request; please retry the operation. For more information on exception types and proper exception handling, please refer to http://go.microsoft.com/fwlink/?LinkId=761101, SystemTracker:NoSystemTracker, Timestamp:2019-01-15T19:03:45

Checked many times, this is the error I see. 
Are we expecting something like "Cannot accept this type of expression" ?

test/topicFilters.spec.ts
HarshaNalluru: It has the rule name. Will add that check too.
This is the name of the error: `MessagingEntityAlreadyExistsError`
No need of console.log(), will remove it.

Will do the same at all the places !

test/topicFilters.spec.ts
HarshaNalluru: Yes, already noted this. I wanted to add something like this.

test/topicFilters.spec.ts
HarshaNalluru: Action is not getting added eventhough I pass it as an argument.
I checked the filters in the samples, they are working fine.
Tried the action part, action has no effect as is not even getting added.

test/topicFilters.spec.ts
HarshaNalluru: Yes, you're right. Removing this.

test/topicFilters.spec.ts
HarshaNalluru: As I've already mentioned, the action statement is not getting added in the rules for the subscription.

test/topicFilters.spec.ts
ramya-rao-a: Can you check which of the 2 is failing? Is it `await subscriptionClient.addRule("Priority_1", "priority = 1");` or `await subscriptionClient.addRule("Priority_2", "1")`;

I believe the first one shouldnt fail.

HarshaNalluru: The error is caused by the second one for sure.

ramya-rao-a:
> It doesn't let me add false as the filter

That is a bug. We should fix it. I have logged #186 to track it. Please pick this bug up after you are done with this PR

> SQL Action thing is not working, Unable to change any property. Action statement is not getting added in the rules for the subscription.

This is a bug too. Logged #187

Issue #173, opened 2019-01-14
Simplify the scheduleMessages function to schedule all msgs at a single later time


ramya-rao-a:
We have the `scheduleMessage` api that allows user to schedule a later time for sending a message.

Correspondingly, we have a `scheduleMessages` which is meant to allow user to schedule multiple messages to be sent at a later time. Here each message can be sent at a different time which forces user to create a new object `ScheduleMessage` which is nothing but another json object with the actual message and the queue time.

The signature is different for `scheduleMessages` when compared to `scheduleMessage` and we now have to export yet another type

A much simpler API is to send all given messages at given time which is what the Go and Python SDKs are doing. This way the signature will also stay similar to `scheduleMessage`

ramya-rao-a:
@bterlson thoughts?

ramya-rao-a:
An even simpler api would be to combine the `scheduleMessage` and `scheduleMessages` to 
`function scheduleMessage(scheduleTime, ...messages)`

bterlson:
I like `scheduleMessages` with a rest parameter for all the messages, that makes a lot of sense.

But it seems like the other API is still needed to support scheduling many messages at different times?

Can you link me to what python/go are doing?

ramya-rao-a:
> support scheduling many messages at different times

I wonder if we should do this at all

Go and Python support multiple message scheduling but not at different times

.Net SDK doesn't support scheduling of multiple messages at all

bterlson:
Sounds like we can assume the functionality isn't super important, so I fully support the simpler design!

ramya-rao-a:
Done with #188

PR #188, opened 2019-01-15
Simplify func signature for scheduling messages

ramya-rao-a:
Related to #173 

This PR updates the function signature for `scheduleMessage` and `scheduleMessages` to

- Use scheduleTime before the messages to be consistent with Go & Python SDK. This will also help in the future if we decide to use [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)
- Update `scheduleMessages` to only support a single schedule time instead of separate ones for each message. 
   - This makes the signature similar to that of `scheduleMessage`
   - Avoids the user having to create a new object "ScheduleMessage"
   - Makes feature consistent with Go & Python SDK
- Removes `ScheduleMessage` from index.ts as it is need not be exported and made visible to the user anymore

### DIFF
--- a/examples/samples/scheduledMessages.ts
+++ b/examples/samples/scheduledMessages.ts
@@ -22,7 +22,7 @@ async function sendMessages(queueClient: QueueClient, data: any[]): Promise<void
     console.log(
       `>>>> Sending message:\t ${message.body}, scheduled for UTC: ${scheduledEnqueueTimeUtc}`
     );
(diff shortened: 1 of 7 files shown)


bterlson:
Looks good to me! But why not a single API scheduleMessage that either has a rest parameter of ScheduleMessage[] or a final parameter typed with ScheduleMessage | ScheduleMessage[].

PR #194, opened 2019-01-16
Boolean filter TestCase - 1

ShivangiReja:
Implemented following  tests for Boolean filter : 

-   A subscription with default filter should receive all messages sent via the topic
-   A subscription with false boolean filter shouldnt receive any messages sent via the topic


test/topicFilterCopy.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+    process.env.DEFAULT_SUBSCRIPTION_NAME
+  );
+  await removeAllRules(subscriptionClient);
+}
+
+async function afterEachTest(): Promise<void> {
+  await namespace.close();
+}
+
+async function sendOrder(order: any): Promise<void> {
```
ramya-rao-a: This function is only being called from `sendOrders()`. Can we merge the 2 functions?

test/topicFilterCopy.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+describe.only("Boolean filters", function(): void {
+  beforeEach(async () => {
+    await beforeEachTest();
+  });
+
+  afterEach(async () => {
+    await afterEachTest();
+  });
+
+  async function addFileterAndReceiveOrders(
```
ramya-rao-a: Filter spelling :)

test/topicFilterCopy.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+    should.equal(rules[0].name, "BooleanFilter");
+
+    await sendOrders();
+    const receivedMsgs = await receiveOrders(client);
+
+    return receivedMsgs;
+  }
+
+  it("Subscription with default filter receives all messages", async function(): Promise<void> {
+    const receivedMsgs = await addFileterAndReceiveOrders(false, defaultSubscriptionClient);
```
ramya-rao-a: The idea of having a `defaultSubscriptionClient` is that we dont add/remove rules from it.

`addFileterAndReceiveOrders` will add a rule.

Right now, your test is passing because the default rule on the default subscription wasnt removed.

test/topicFilterCopy.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+    void
+  > {
+    const receivedMsgs = await addFileterAndReceiveOrders(false, subscriptionClient);
+
+    should.equal(Array.isArray(receivedMsgs), true);
+    should.equal(receivedMsgs.length, 0);
+
+    await testPeekMsgsLength(subscriptionClient, 0);
+
+    await receiveOrders(defaultSubscriptionClient);
```
ramya-rao-a: Why are we receiving from default subscription client here?

test/topicFilterCopy.spec.ts
ShivangiReja: Already changed.

test/topicFilterCopy.spec.ts
ShivangiReja: Can I remove defaultSubscriptionClient ?

ShivangiReja: I added that because of this test case `A subscription with default filter should receive all messages sent via the topic`.

test/topicFilterCopy.spec.ts
ShivangiReja: Calling `receiveOrders(defaultSubscriptionClient);` just for the cleanup.

test/topicFilters.spec.ts
```diff
@@ -61,6 +83,57 @@ async function afterEachTest(): Promise<void> {
...
+    (msg: ServiceBusMessage) => {
+      receivedMsgs.push(msg);
+      return Promise.resolve();
+    },
+    (err: Error) => {
+      should.not.exist(err);
+    }
+  );
+
+  await delay(3000);
```
ramya-rao-a: Lets update this to 5 seconds. Sometimes when I have run the streaming receiver, some messages never came in until I waited for 5 seconds

ramya-rao-a: Lets update this to 5 seconds. Sometimes when I have run the streaming receiver, some messages never came in until I waited for 5 seconds

test/topicFilters.spec.ts
```diff
@@ -61,6 +83,57 @@ async function afterEachTest(): Promise<void> {
...
+    { Color: "blue", Quantity: 10, Priority: "low" },
+    { Color: "blue", Quantity: 5, Priority: "high" },
+    { Color: "blue", Quantity: 10, Priority: "low" },
+    { Color: "red", Quantity: 5, Priority: "low" },
+    { Color: "red", Quantity: 10, Priority: "low" },
+    { Color: "red", Quantity: 5, Priority: "low" },
+    { Color: "yellow", Quantity: 10, Priority: "high" },
+    { Color: "yellow", Quantity: 5, Priority: "low" },
+    { Color: "yellow", Quantity: 10, Priority: "low" }
+  ];
```
ramya-rao-a: Lets move this out of this function, so that we can then use `data.length` instead of the hard-coded number 12 in the tests

PR #202, opened 2019-01-16
Tests for Send using sessions

ramya-rao-a:
This PR adds tests for the below scenarios for partitioned/unpartitioned queues/topics using sessions
- simple send
- schedule message
- schedule multiple messages
- cancel a scheduled message
- cancel multiple scheduled messages

### DIFF
--- a/test/send.spec.ts
+++ b/test/send.spec.ts
@@ -10,39 +10,22 @@ chai.use(chaiAsPromised);
 import {
   Namespace,
   QueueClient,
-  SendableMessageInfo,
-  generateUuid,
   TopicClient,
   SubscriptionClient,
-  delay
+  delay,
+  MessageSession
 } from "../lib";
 
-const testMessages: SendableMessageInfo[] = [
-  {
-    body: "hello1",
-    messageId: `test message ${generateUuid()}`
-  },
-  {
-    body: "hello2",
-    messageId: `test message ${generateUuid()}`
-  }
-];
-
-const testMessagesToSamePartitions: SendableMessageInfo[] = [
-  {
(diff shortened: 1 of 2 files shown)


test/send.spec.ts
```diff
@@ -157,22 +228,62 @@ describe("Send to Queue/Subscription", function(): void {
...
+      partitionedSubscriptionMessageSession,
+      true
+    );
+  });
+
+  it.only("Simple send using unPartitioned Queues with Sessions", async function(): Promise<void> {
+    await testSimpleSend(unpartitionedQueueSessionClient, unpartitionedQueueMessageSession, true);
+  });
+
+  it.only("Simple send using unPartitioned Topics and Subscriptions with Sessions", async function(): Promise<
```
ShivangiReja: Why are we adding "Only" here and in other "it" and "describe" section?

test/send.spec.ts
```diff
@@ -278,34 +473,93 @@ describe("Cancel Scheduled messages for sending to Queue/Subscription", function
...
+    void
+  > {
+    await testCancelScheduleMessage(
+      partitionedTopicSessionClient,
+      partitionedSubscriptionMessageSession,
+      true
+    );
+  });
+
+  it("Cancel a single Scheduled message using unPartitioned Queues with Sessions", async function(): Promise<
```
ShivangiReja: P (`unPartitioned` ) should be lower case.

test/send.spec.ts
ramya-rao-a: We dont need that, I forgot to remove it after testing that test

Issue #43, opened 2018-12-11
Standardize the structure for samples


ramya-rao-a:
Samples currently live at https://github.com/Azure/azure-service-bus-node/tree/master/examples/samples

We need to come up with a structure for our samples such that all samples are consistent, concise and to the point when demonstrating the feature

Some things to keep in mind:

- How to fetch the connection string and entity names?
   - We are currently using the `dotenv` module which expects the `.env` file to reside in the folder from where you are running the samples. This will fail if the user has any other working directory
- Error handling
   - Do we let the errors bubble up so that if we ever hook this up with CI, we will know that the samples are failing?
   - If not, then we need to have a consistent way of using `try/catch` and `.catch()`
- Ensure namespace is closed in both the happy path scenario and the error scenarios
- Naming conventions for files and functions... pascal case vs camel case
- What/what not and When/When not to console.log 
- Structure for console.logs ### vs <<< vs >>>
- All samples should have a comment block up top that would have set up instructions if any along with what the sample demonstrates, plus any helpful links for further reading

cc @AlexGhiondea @ShivangiReja

ramya-rao-a:
Now that we have more samples coming in, here are a few more notes:

- `timeToLive`: Test if this indeed removes messages from the queue or just moves it to the deadletter queue. If we intend to use this so that we dont have to perform a clean up later, then the comment on it should say the same. Else it gives an impression that it is a mandatory property
- `messageId`: Most of our samples dont really use this property. Why have it?
- Right now, each sample has code to send the messages, though most samples are about receiving in different ways. If we remove the sending part from the sample, the sample would be more concise and to the point. In the setup instructions, we can point to the sample that is meant just for sending messages in case the user is working with a non-empty queue

ramya-rao-a:
cc @HarshaNalluru @ramya0820

ramya-rao-a:
- Some samples will not work as expected unless the queue is empty. For example: The `receiveLoop.ts` sends 10 messages and receives 10 messages. If the queue is not empty, then you wont get the messages that you actually sent
- Some samples demonstrate features of Service Bus, but not necessarily of the SDK. Examples: autoForward.ts and duplicateDetection.ts. The samples dont show anything new from the SDK's perspective. Its send and receive. What is the point of these samples?

PR #204, opened 2019-01-16
Refactor samples v2

ramya0820:
## Description

- Refactored to
      ___ examples
      ______ getting started
         ______ advanced features
          ______ to be deleted // Contains older samples for reference and review
               
- Added summaries for the examples
- Made examples queue/topics agnostic and added comments
- Added in notes to re-use sendMessages sample
- Fixed/uniformized error handling, message attributes in use, and few other clean ups as discussed on #43 and in person

- Testing:- Individually executed samples, found and filed bug reports

- #43


examples/gettingStarted/scheduledMessages.ts
```diff
@@ -0,0 +1,104 @@
...
+  try {
+    await sendScheduledMessages();
+
+    await receiveMessages();
+  } finally {
+    await ns.close();
+  }
+}
+
+// Scheduling messages to be sent after 30 seconds from now
```
ramya-rao-a: Comment says 30 seconds, but we are using 10 seconds

examples/gettingStarted/scheduledMessages.ts
```diff
@@ -0,0 +1,104 @@
...
+  const timeNowUtc = new Date(Date.now());
+  const scheduledEnqueueTimeUtc = new Date(Date.now() + 10000);
+  console.log(`>>>> Time now in UTC: ${timeNowUtc}`);
+  console.log(`>>>> Messages will appear in Service Bus at UTC: ${scheduledEnqueueTimeUtc}`);
+
+  await client.scheduleMessages(scheduledEnqueueTimeUtc, messages);
+  for (let index = 0; index < messages.length; index++) {
+    await client.scheduleMessage(scheduledEnqueueTimeUtc, messages[index]);
+    console.log(`Sent: ${messages[index].body} - ${messages[index].label}`);
+  }
```
ramya-rao-a: Is the intention here to demonstrate both `scheduleMessage` and `scheduleMessages`?
I would prefer to keep the sample simpler, and just do one of the 2.

examples/gettingStarted/scheduledMessages.ts
```diff
@@ -0,0 +1,104 @@
...
+  const onError: OnError = (err: MessagingError | Error) => {
+    console.log("Error occurred: ", err);
+  };
+
+  let numOfMessagesReceived = 0;
+  const receiveListenerFirst = client.receive(onMessage, onError);
+  await delay(5000);
+  await receiveListenerFirst.stop();
+  console.log(
+    `\nStarted looking up immediately while having receive listener wait for 5 seconds. Found ${numOfMessagesReceived} messages.\n`
```
ramya-rao-a: Any specific reason for including the newlines? Each `console.log` would show up in a new line automatically.

`${numOfMessagesReceived} messages found after waiting for 5 seconds` is much simpler

Same for the other console.log

examples/gettingStarted/scheduledMessages.ts
```diff
@@ -0,0 +1,104 @@
...
+  );
+
+  await delay(5000);
+
+  const receiveListenerSecond = client.receive(onMessage, onError);
+  await delay(5000);
+  await receiveListenerSecond.stop();
+  console.log(
+    `\nStarted looking up after 10 seconds while having receive listener wait, received ${numOfMessagesReceived} messages.`
+  );
```
ramya-rao-a: By this time, we have waited for a total of 15 seconds not 10.

examples/gettingStarted/scheduledMessages.ts
ramya0820: The output looks neater, less confusing I think

examples/gettingStarted/scheduledMessages.ts
ramya0820: Technically, lookup started after 10 seconds. The messages start getting retrieved/displayed at 10 seconds. Lookup finished at 15 seconds

ramya-rao-a: I am guessing the intention of having the 2 receivers is to show that by the time the first message is printed, no messages were received. By the time the second one was printed, all (or atleast non zero) messages were received

When the user runs the sample and sees the messages, the point of reference they have is the scheduling of the messages.
The fact that we have 2 listeners or how much they waited is inconsequential to them until they go line by line in the code.

The first message is printed 5 seconds after scheduling the messages.
The second message is printed 15 seconds after scheduling the messages.

I believe it would be simpler if the messages used the scheduling of messages as point of reference.

examples/gettingStarted/scheduledMessages.ts
```diff
@@ -0,0 +1,103 @@
...
+  }));
+
+  const timeNowUtc = new Date(Date.now());
+  const scheduledEnqueueTimeUtc = new Date(Date.now() + 10000);
+  console.log(`>>>> Time now in UTC: ${timeNowUtc}`);
+  console.log(`>>>> Messages will appear in Service Bus at UTC: ${scheduledEnqueueTimeUtc}`);
+
+  await client.scheduleMessages(scheduledEnqueueTimeUtc, messages);
+  for (let index = 0; index < messages.length; index++) {
+    console.log(`Sent: ${messages[index].body} - ${messages[index].label}`);
```
ramya-rao-a: Do we need to print this here? I am guessing this is a remnant of the time when we were scheduling individual messages

examples/gettingStarted/scheduledMessages.ts
```diff
@@ -0,0 +1,103 @@
...
+  { name: "Einstein", firstName: "Albert" },
+  { name: "Heisenberg", firstName: "Werner" },
+  { name: "Curie", firstName: "Marie" },
+  { name: "Hawking", firstName: "Steven" },
+  { name: "Newton", firstName: "Isaac" },
+  { name: "Bohr", firstName: "Niels" },
+  { name: "Faraday", firstName: "Michael" },
+  { name: "Galilei", firstName: "Galileo" },
+  { name: "Kepler", firstName: "Johannes" },
+  { name: "Kopernikus", firstName: "Nikolaus" }
```
ramya-rao-a: I was wondering if we are gaining anything by having the first and last name split like this and then joining them together when creating the message body.

Why not just have `listOfScientists` be an array of string?

examples/gettingStarted/scheduledMessages.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import {
+  Namespace,
+  SendableMessageInfo,
+  OnMessage,
+  MessagingError,
+  OnError,
+  ServiceBusMessage
+} from "../../lib";
+import { delay } from "rhea-promise";
```
ramya-rao-a: This would need the user to have `rhea-promise` as a dependency, which wouldnt be ideal when we want to enable them to just copy/paste the code.

We should figure out an alternative

examples/gettingStarted/session.ts
```diff
@@ -0,0 +1,86 @@
...
+async function sendMessages(sessionId: string): Promise<void> {
+  // If using Topics, use createTopicClient to send to a topic
+  const client = ns.createQueueClient(queueName);
+  const data = [
+    { step: 1, title: "Shop" },
+    { step: 2, title: "Unpack" },
+    { step: 3, title: "Prepare" },
+    { step: 4, title: "Cook" },
+    { step: 5, title: "Eat" }
+  ];
```
ramya-rao-a: I believe the reason we are using steps here is so that we can then show that sessions has the FIFO feature.

I wonder if as a SDK sample, it is the responsibility of this sample to demonstrate FIFO. This uses the same argument as we used for not having the duplicateDetection sample.

The main responsibility of this sample would be just to show how to send/receive messages.

Thoughts?

If you agree, then I would fall back on using the simple scientists data that we are using in the other samples

ramya-rao-a:
@ramya0820 I have pushed a commit to this branch with the following changes

The samples in the advanced section are updated to
  - Move the readme comment up top
  - Remove console.logging of connection string, queue name etc.
  - Remove dotenv module usage

Other changes:
- Include links & setup info (if applicable) in the readme
- Skip type imports if possible (SendableMessageInfo, ServiceMessage, MessagingError, Error) 
- Avoid modular approach when there is just 1 other function other than `main()`
- Print just the body (not label), when the sample doesnt include the sending part.
- Avoid having namespace as a global variable. This was previously needed when we used to close the namespace only after calling main()
- Deferral: Use `scheduledEnqueueTimeUtc` property to send messages out of order. This simplifies the code a bit
- Topic Filters: Updated as per comments on the previous PR

Please pull from this branch before your next set of changes and let me know if you have comments for any of the above.

examples/gettingStarted/scheduledMessages.ts
ramya0820: Yes, agree with using scheduled time as reference, which is 10 seconds. So mentioning 15 seconds may not seem relevant.
Updated initial message to indicate the 10 second scheduling delay that the sample is using.

examples/gettingStarted/scheduledMessages.ts
ramya0820: I had intentionally left that in for clarity on output messages

examples/gettingStarted/scheduledMessages.ts
ramya0820: Maybe using JSON representation is a little more closer to real world examples? Either way, I'm okay with to update.
Also, just noticed the attributes could be better named, updated these.

examples/gettingStarted/session.ts
ramya0820: Hm.. agree. Since the usability study meeting, I do think it makes sense to have copy-paste-able code handy in the samples. Will simplify

examples/gettingStarted/scheduledMessages.ts
ramya0820: Will address this in a different issue

Issue #193, opened 2019-01-16
Issue with Try Catch handling in SDK (messageReceiver.ts)


ramya0820:
## Description:
Errors thrown by OnMessage callback are being eaten by SDK, need to expose them back to the user.

## Context:
Found that the assertions written within callbacks are not being executed in the tests.

ramya-rao-a:
Fixed with #207

PR #207, opened 2019-01-17
Error from message handler should surface up

ramya-rao-a:
## Description

This PR fixes the issue described in #193 

Simply throwing the error back after calling the user's message handler doesnt let the error reach the user. This is because the error gets bubbled up to the code that calls the message handler, and this code is way deep in the call stack. 

So instead, this PR resorts to call the user's error handler.

cc @HarshaNalluru for the changes to the tests for the streaming receiver


ramya-rao-a:
This is pending investigation on how such case is handled in the .Net SDK which @HarshaNalluru is looking into

HarshaNalluru:
> This is pending investigation on how such case is handled in the .Net SDK which @HarshaNalluru is looking into

[BasicSendReceiveQuickStart - .NET sample](https://github.com/Azure/azure-service-bus/blob/master/samples/DotNet/GettingStarted/BasicSendReceiveQuickStart/BasicSendReceiveQuickStart/Program.cs)

[`ExceptionReceivedHandler`](https://github.com/Azure/azure-service-bus/blob/94f7a0540dc6d0a5896f97a1e7e6ff1866705390/samples/DotNet/GettingStarted/BasicSendReceiveQuickStart/BasicSendReceiveQuickStart/Program.cs#L95)(user's error handler) is getting called when an error is thrown from [`ProcessMessagesAsync`](https://github.com/Azure/azure-service-bus/blob/94f7a0540dc6d0a5896f97a1e7e6ff1866705390/samples/DotNet/GettingStarted/BasicSendReceiveQuickStart/BasicSendReceiveQuickStart/Program.cs#L80)(user's message handler)

ramya-rao-a:
Thanks @HarshaNalluru!

So, with this change, we would doing the same as the .Net SDK

test/receiveAndDeleteMode.spec.ts
```diff
@@ -136,7 +136,9 @@ describe("ReceiveBatch from Queue/Subscription", function(): void {
 });
 
 describe("Streaming Receiver from Queue/Subscription", function(): void {
+  let errorFromErrorHandler: Error | undefined;
```
ramya0820: Overall found the test code confusing/hard to follow at first glance
Can rename to "uncaughtErrorFromHandlers"

test/receiveAndDeleteMode.spec.ts
```diff
@@ -157,7 +159,9 @@ describe("Streaming Receiver from Queue/Subscription", function(): void {
         return Promise.resolve();
       },
       (err: Error) => {
-        should.not.exist(err);
+        if (err) {
```
ramya0820: Do we need the if check? The callback would executes only if there was an error.

Can replace with:

uncaughtErrorFromHandlers = err;

test/receiveAndDeleteMode.spec.ts
```diff
@@ -171,6 +175,9 @@ describe("Streaming Receiver from Queue/Subscription", function(): void {
     should.equal(receivedMsgs[0].messageId, testMessages[0].messageId);
 
     await receiveListener.stop();
+    if (errorFromErrorHandler) {
+      should.fail(!!errorFromErrorHandler, false, errorFromErrorHandler.message);
```
ramya0820: The intent seems to be to simply fail the test if there's an error.

We can use chai.assert.fail(uncaughtErrorFromHandlers.message) and avoid the true, false comparison (and potential bugs from interpreting what the fail is supposed to do).

should.fail() seems generally confusing, I was expecting it to have just one input param - the failing message (like how assert.fail does) but *shrugs*

test/receiveAndDeleteMode.spec.ts
ramya0820: Even better, I think the error handling code is to just make sure the tests are all working correctly.
Technically it never gets invoked.
We can define a common onError handler for tests and move the verification check ti .afterEach() block. 
This way the tests are much cleaner and we don't need to add much code.

@ramya-rao-a @HarshaNalluru

ramya-rao-a: Good point.
Updated to have a common error handler.

But I havent moved the check on whether the error was thrown or not to afterEach() block.
There are currently checks after this check and before afterEach().
In the event where the unexpected error is thrown, these checks will fail as well.
If the unexpected error check moves to the afterEach() block, then it will never be run as the above checks will fail

test/receiveAndDeleteMode.spec.ts
ramya-rao-a: Updated this check inside the `assert.fail`, so we dont need this anymore.
We had this to get around the linter error which tells that err can possibly be undefined

test/receiveAndDeleteMode.spec.ts
ramya-rao-a: Good point! Updated to use assert.fail()

test/receiveAndDeleteMode.spec.ts
ramya0820: At a time only one error will be caught. Although ideally sequence of error catching is ideal like you have mentioned. I still think it's cleaner to have it in afterEach(). Doesnt seem incorrect, can sync up offline later. Not a blocker (Although I'm trying to better understand how to best structure the tests in general!)

ramya-rao-a: The major driving point for me is the scenario where 6 months down the line a change in the SDK breaks the test. If the sequence is not right, that engineer might end up wasting half the day chasing the wrong errors.

ramya-rao-a: Also, now that the check is just 1 line `chai.assert.fail()`, I dont think we have the issue of things not being clean anymore

test/streamingReceiver.spec.ts
```diff
@@ -184,6 +188,9 @@ describe("Streaming Receiver Misc Tests", function(): void {
     }
 
     await receiveListener.stop();
+
+    chai.assert.fail(unexpectedError && unexpectedError.message);
```
HarshaNalluru: @ramya-rao-a 
This check(`chai.assert.fail`) is throwing the error even if the test passes(`unexpectedError = undefined`).
> AutoComplete removes the message from Partitioned Queues(with sessions):
     `AssertionError: assert.fail()`

ramya-rao-a: @HarshaNalluru That's right. This got fixed in https://github.com/Azure/azure-service-bus-node/commit/9e3c868557f17d1f751e403ee3553cf129ec63bb

But looks like there are 2 other files with this. I'll send a PR for that

ramya-rao-a: @HarshaNalluru Sent PR #217. If you take a look and approve, then I'll merge it

PR #208, opened 2019-01-17
Test cases for Topic Filters.

ShivangiReja:
Implemented following tests:
SQL filters:
- SQL rule filter on one of the message properties.
- SQL rule filter on one of the custom properties.
- SQL rule filter using the AND operator.
- SQL rule filter using the OR operator.
- SQL rule filter and an action.

Correlation filters:
- Correlation filter on the message properties.
- Correlation filter on the custom properties.
- Correlation filter and a SQL action.

### DIFF
--- a/test/topicFilters.spec.ts
+++ b/test/topicFilters.spec.ts
@@ -76,6 +76,16 @@ async function beforeEachTest(): Promise<void> {
     process.env.TOPIC_NAME,
     process.env.DEFAULT_SUBSCRIPTION_NAME
   );
+
+  const peekedDefaultSubscriptionMsg = await defaultSubscriptionClient.peek();
+  if (peekedDefaultSubscriptionMsg.length) {
+    throw new Error("Please use an empty Default Subscription for integration testing");
+  }
+
+  const peekedSubscriptionMsg = await subscriptionClient.peek();
+  if (peekedSubscriptionMsg.length) {
+    throw new Error("Please use an empty Subscription for integration testing");
+  }
   await removeAllRules(subscriptionClient);
 }
 
@@ -108,7 +118,7 @@ async function sendOrders(): Promise<void> {
       label: `${element.Color}`,
       userProperties: {
         color: `${element.Color}`,
-        quantity: `${element.Quantity}`,
+        quantity: element.Quantity,
         priority: `${element.Priority}`
       }
     };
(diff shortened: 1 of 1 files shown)


test/topicFilters.spec.ts
```diff
@@ -431,3 +441,171 @@ describe("Send/Receive messages using boolean filters of subscription", function
...
+      "SET priority = 'High'"
+    );
+
+    const rules = await subscriptionClient.getRules();
+    should.equal(rules.length, 1, "rule");
+    should.equal(rules[0].name, "CorrelationRuleWithAction");
+
+    await sendOrders();
+    const receivedMsgs = await receiveOrders(subscriptionClient);
+
```
HarshaNalluru: Can you add one more check to see whether the `action` actually updates the attribute ?

test/topicFilters.spec.ts
```diff
@@ -431,3 +441,190 @@ describe("Send/Receive messages using boolean filters of subscription", function
...
+    const rules = await subscriptionClient.getRules();
+    should.equal(rules.length, 1, "rule");
+    should.equal(rules[0].name, "SQLMsgPropertyRule");
+
+    await sendOrders();
+    const receivedMsgs = await receiveOrders(subscriptionClient);
+
+    should.equal(Array.isArray(receivedMsgs), true);
+    should.equal(receivedMsgs.length, 4);
```
ramya-rao-a: If we ever change the input data, then we will have to go through all the tests and update this number. Use something like `data.filter(msg => msg.label == "red").length` instead

Same for other tests

test/topicFilters.spec.ts
```diff
@@ -431,3 +441,190 @@ describe("Send/Receive messages using boolean filters of subscription", function
...
+
+    await testPeekMsgsLength(subscriptionClient, 0);
+  });
+
+  it("SQL rule filter on the custom properties", async function(): Promise<void> {
+    await subscriptionClient.addRule("SQLCustomRule", "color='red'");
+
+    const rules = await subscriptionClient.getRules();
+    should.equal(rules.length, 1, "rule");
+    should.equal(rules[0].name, "SQLCustomRule");
```
ramya-rao-a: Maybe we can add a function that does the adding rule and doing the above 2 checks? This part is repeated in a few test cases

test/topicFilters.spec.ts
```diff
@@ -431,3 +441,190 @@ describe("Send/Receive messages using boolean filters of subscription", function
...
+
+    await afterEachTest();
+  });
+
+  it("SQL rule filter on the message properties", async function(): Promise<void> {
+    await subscriptionClient.addRule("SQLMsgPropertyRule", "sys.label = 'red'");
+
+    const rules = await subscriptionClient.getRules();
+    should.equal(rules.length, 1, "rule");
+    should.equal(rules[0].name, "SQLMsgPropertyRule");
```
ramya-rao-a: We are adding new rules to the subscription in each test case. They all get removed before started a new test, but by the end, we are left with the rule added by the last test case.

We should update the `afterEach` section to remove all the filters and then add the default filter back. Because, in the absence of any rules, the subscription wont get any messages. This can fail tests in the other files

Issue #201, opened 2019-01-16
There are some warnings, when running npm run units.


ShivangiReja:
**Describe the bug**

dist-esm/test/**/*.spec.js → test-dist/index.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en#warning-treating-module-as-external-dependency
fs (imported by node_modules\dotenv\lib\main.js,  commonjs-external:fs)
path (imported by node_modules\dotenv\lib\main.js,  commonjs-external:path)
(!) Circular dependency: node_modules\chai\lib\chai.js -> node_modules\chai\lib\chai\utils\index.js -> node_modules\chai\lib\chai\utils\addProperty.js -> node_modules\chai\lib\chai.js
(!) Circular dependency: node_modules\chai\lib\chai.js -> node_modules\chai\lib\chai\utils\index.js -> node_modules\chai\lib\chai\utils\addProperty.js ->  commonjs-proxy:C:\Users\shreja\azure-service-bus-node\node_modules\chai\lib\chai.js -> node_modules\chai\lib\chai.js
(!) Circular dependency: node_modules\chai\lib\chai.js -> node_modules\chai\lib\chai\utils\index.js -> node_modules\chai\lib\chai\utils\addMethod.js -> node_modules\chai\lib\chai.js
(!) Circular dependency: node_modules\chai\lib\chai.js -> node_modules\chai\lib\chai\utils\index.js -> node_modules\chai\lib\chai\utils\overwriteProperty.js -> node_modules\chai\lib\chai.js
(!) Circular dependency: node_modules\chai\lib\chai.js -> node_modules\chai\lib\chai\utils\index.js -> node_modules\chai\lib\chai\utils\overwriteMethod.js -> node_modules\chai\lib\chai.js
(!) Circular dependency: node_modules\chai\lib\chai.js -> node_modules\chai\lib\chai\utils\index.js -> node_modules\chai\lib\chai\utils\addChainableMethod.js -> node_modules\chai\lib\chai.js
(!) Circular dependency: node_modules\chai\lib\chai.js -> node_modules\chai\lib\chai\utils\index.js -> node_modules\chai\lib\chai\utils\overwriteChainableMethod.js -> node_modules\chai\lib\chai.js
created test-dist/index.js in 1.3s

ramya-rao-a:
Fixed with #209

Issue #196, opened 2019-01-16
After moving to the new build system there are warnings when running npm run unit


AlexGhiondea:
The error looks like this:

```
73: describe("ReceiveBatch from Queue/Subscription", () => {
74:     beforeEach(() => tslib_1.__awaiter(this, void 0, void 0, function* () {
                                           ^
75:         yield beforeEachTest();
76:     }));
```

PR #209, opened 2019-01-17
Fix test rollup

bterlson:
fs and path are required by dotenv, so need to be declared as external.

The this-is-undefined errors are a result of TypeScript compilation and can be safely ignored.

The circular reference errors are unfortunately real, but don't seem to impact the rollup (IIUC, rollup picks some ordering which may not work properly given the circularity, but depending on usage, the circularity might not be observable). So this PR ignores those errors as well.


bterlson:
Fixes #201. /cc @ramya-rao-a @ShivangiReja

ramya-rao-a:
@btardif Where are you seeing the "this is undefined" error?
Is that related to https://github.com/Azure/azure-service-bus-node/issues/196?

AlexGhiondea:
@bterlson I think I fixed the 'this-is-undefined' errors already.

Thanks for fixing the remaining rollup issues!!!

bterlson:
I was getting the error with latest master. To confirm: the `this is undefined` error is fixed in master?

bterlson:
Ok I think I know what's happening. I see that @AlexGhiondea did fix this as of 1066e2d1b, however at least e146780815e6966dfc5d0fb5d9f8d73228ad1a93 reintroduced code that shows this pattern. IMO 1066e2d1b should be reverted in favor of the ignore in this commit. It's not really realistic to avoid all patterns that result with `this` at global scope, and I really don't want to write less readable code just to make rollup happy.

Issue #195, opened 2019-01-16
Implement send/batching/streaming receiver tests for Sessions


ramya-rao-a:
- Repeat the tests (other than the auto lock renewal and the reconnecting ones) in #100 and #101 for a single session after using `acceptSession()`

ramya-rao-a:
Send tests with sessions are done with #202

PR #211, opened 2019-01-18
Tests - Batching Receiver - with and without sessions

HarshaNalluru:
## Description
Updated the tests in Batching Receiver to run the tests for the Partitioned and Unpartitioned Queues/Subscriptions with and without sessions

- #195


HarshaNalluru:
**One observation**
While executing the tests for `Throws error when call the second ReceiveBatch while the first one is not done`, I realized the two modules 
> `testParallelReceiveBatchCalls`(from **Multiple ReceiveBatch calls**) and 
> `simulatenousReceiveBatch`(from **Batching Receiver Misc Tests**) 

have the exact same code !
```typescript
 
  // We use an empty queue/topic here so that the first receiveBatch call takes time to return
  async function testParallelReceiveBatchCalls(
    receiverClient: QueueClient | SubscriptionClient | MessageSession
  ): Promise<void> {
    const firstBatchPromise = receiverClient.receiveBatch(1, 10);
    await delay(5000);
    const secondBatchPromise = receiverClient.receiveBatch(1, 10).catch((err) => {
      should.equal(err.name, "Error");
      errorWasThrown = true;
    });
    await Promise.all([firstBatchPromise, secondBatchPromise]);
    should.equal(errorWasThrown, true);
  }
```
```typescript
 
  async function simulatenousReceiveBatch(
    receiverClient: QueueClient | SubscriptionClient | MessageSession
  ): Promise<void> {
    const firstBatchPromise = receiverClient.receiveBatch(1, 10);
    await delay(5000);
    const secondBatchPromise = receiverClient.receiveBatch(1, 10).catch((err) => {
      should.equal(err.name, "Error");
      errorWasThrown = true;
    });
    await Promise.all([firstBatchPromise, secondBatchPromise]);
    should.equal(errorWasThrown, true);
  }
```
So, basically the last eight tests are redundant and can be removed.
We should still be figuring out why "**The test case rightly points out the error but the test still fails.**" for the session enabled ones [`Throws error when call the second ReceiveBatch while the first one is not done`]

@ramya-rao-a, Once you confirm that they can be removed I'll remove the last eight tests and will push the changes.

ramya-rao-a:
Good catch on the redundant tests :) I have pushed a commit to remove them.

I have also added tests for unpartitioned, session enabled queues/subscriptions for deferred cases. These are the ones that we previously had commented tests for the non session cases. I just added commented tests for sessions as well.

Updated the tests for the deadletter queues to use `testSimpleMessages` directly instead of passing it around as a parameter.

I'll start running the tests that failed for you now.

ramya-rao-a:
@HarshaNalluru 

- The InternalServerError when trying to settle a deferred message is being tracked in https://github.com/Azure/azure-service-bus-node/issues/212
- The test failure for "Multiple Receive Batch" calls is fixed with https://github.com/Azure/azure-service-bus-node/commit/9e3c868557f17d1f751e403ee3553cf129ec63bb
- The timeouts that you see for maxDeliveryCount, I dont see it. So, thats fine.
- The MessageTimeout error for `testAskForMore` is being tracked in #216
- The last set was the redundant tests

PR #215, opened 2019-01-18
Fix failing tests

ramya-rao-a:
## Description

This PR fixes 2 kinds of failing tests

- The ones in maxDeliveryCount.spec.ts were failing due to the `chai.assert.fail` that would fail all the time. Replaced this with a `should` check
- Updated receiveBatch() function in the messageSession.ts file to be async. This fixes the failures of "Throws error when ReceiveBatch is called while the previous call is not done" test cases for sessions as mentioned in https://github.com/Azure/azure-service-bus-node/pull/211#issuecomment-455451970

### DIFF
--- a/lib/session/messageSession.ts
+++ b/lib/session/messageSession.ts
@@ -625,7 +625,7 @@ export class MessageSession extends LinkEntity {
    * 10% of the lockDuration value.
    * - **Default**: `2` seconds.
    */
-  receiveBatch(
+  async receiveBatch(
     maxMessageCount: number,
     maxWaitTimeInSeconds?: number,
     maxMessageWaitTimeoutInSeconds?: number
--- a/test/maxDeliveryCount.spec.ts
+++ b/test/maxDeliveryCount.spec.ts
@@ -290,7 +290,7 @@ describe("Streaming Receiver: Message abandoned more than maxDeliveryCount goes 
     await delay(4000);
 
     await receiveListener.stop();
-    chai.assert.fail(unexpectedError && unexpectedError.message);
+    should.equal(unexpectedError, undefined, unexpectedError && unexpectedError.message);
 
     should.equal(checkDeliveryCount0, maxDeliveryCount);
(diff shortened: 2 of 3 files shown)


lib/session/messageSession.ts
```diff
@@ -625,7 +625,7 @@ export class MessageSession extends LinkEntity {
    * 10% of the lockDuration value.
    * - **Default**: `2` seconds.
    */
-  receiveBatch(
+  async receiveBatch(
```
AlexGhiondea: Do we need to change the places where we are calling this?

ramya-rao-a: Its only being called in the test cases. And no the way it is called doesnt need to change. A function that returns a promise can be called with await even if it hasnt been marked as async

ramya-rao-a: The problem here was that since it wasnt marked as async, errors thrown from inside would only be caught in a try/catch

Marking it as async lets us attach a .catch() to the call

bterlson: If it helps, I often explain it thusly: from the outside, there is very little observable semantic difference between

```
async function foo () { ... }`
```

and

```
function foo() {
  return new Promise(() => { ... });
}
```

Users can await using async functions or call and use promise methods, either works fine. Further, it is not a breaking change to refactor a promise-returning function that looks like above into an async function.

In this case, the fix is to bring the code in line with either pattern. Making the function async is one way to do it. Another would be to move the `new Promise()` part to the top of the function so that the error checking stuff happens inside the promise constructor callback (so any errors thrown will reject the promise rather than bubble out synchronously).

ramya0820:
I just noticed this PR - assert.fail() is meant to always fail if hit. It's generally used to mark code pathways that shouldn't be entered/executed as part of the test-case.
E.g usage -
if (uncaughtErrorFromHandlers) {
    chai.assert.fail(uncaughtErrorFromHandlers.message);
  }

ramya-rao-a:
@ramya0820 Yes, thats right. We fixed it with another PR. Check the master branch

PR #217, opened 2019-01-19
Fix other test failures due to assert.fail

ramya-rao-a:
## Description

Follow up to https://github.com/Azure/azure-service-bus-node/pull/215 to fix the chai.assert.fail() cases


ramya-rao-a:
@HarshaNalluru Fixed with https://github.com/Azure/azure-service-bus-node/pull/217/commits/56a4e5071202587d075534a5dd03f3f3d6ad883c

I had seen this when trying out the tests, but had earlier updated only in one of the places :(

Now, things should be better.

PR #221, opened 2019-01-20
Testing DeliveryCount for defer() and abandon()

HarshaNalluru:
## Description
Testing DeliveryCount for `defer()` and `abandon()` in [`streamingReceiver.spec.ts`](https://github.com/Azure/azure-service-bus-node/blob/master/test/streamingReceiver.spec.ts)
- Tests corresponding to `defer()` didn't fail 
image

- Tests corresponding to `abandon()` fail sometimes.
   - Observed `DeliveryCount` is not the same as expected `DeliveryCount`.
   - **Cause** - Not being able to stop the `receiveListener` before the invocation of second `receive()`(This makes `DeliveryCount` equal to `2` instead of `1` ).

- https://github.com/Azure/azure-service-bus-node/issues/101#issuecomment-454153613

### DIFF
--- a/test/streamingReceiver.spec.ts
+++ b/test/streamingReceiver.spec.ts
@@ -494,9 +494,12 @@ describe("Defer message", function(): void {
     }
     should.equal(deferredMsg0.body, testMessages[0].body);
     should.equal(deferredMsg0.messageId, testMessages[0].messageId);
+    should.equal(deferredMsg0.deliveryCount, 1);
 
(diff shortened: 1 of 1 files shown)


HarshaNalluru:
Raised an issue for the [abandon()](https://github.com/Azure/azure-service-bus-node/blob/ef327b7d3fd7c5e1ac88629123618ac945ce4581/test/streamingReceiver.spec.ts#L376) at #222

Issue #212, opened 2019-01-18
Error when settling a deferred message in a session enabled queue/topic


ramya-rao-a:
When writing test cases for settling a deferred message from a session enabled queue/topic, we are getting an InternalServerError

See https://github.com/Azure/azure-service-bus-node/pull/211#issuecomment-455451970

![image](https://user-images.githubusercontent.com/16890566/51406956-769d7200-1b0f-11e9-9920-8c86c167a9c4.png)

Also, once this error is thrown, the program doesnt exit. Instead it looks like it is stuck doing something which then results in lock errrors:

![image](https://user-images.githubusercontent.com/16890566/51407041-b2d0d280-1b0f-11e9-9959-410d5775d1c0.png)

ramya-rao-a:
Fixed with https://github.com/Azure/azure-sdk-for-js/pull/972

PR #223, opened 2019-01-21
Streaming Receiver : Partitioned/Unpartitioned - with sessions

HarshaNalluru:
## Description
Testing sessions in Streaming Receiver - Partitioned and Unpartitioned Queues/Subscriptions

- #195


HarshaNalluru:
@ramya-rao-a 
**Observation [for the failed cases of "Abandon message(with sessions)"]** 

`receiveHandler.stop()` can't be called 
- receive() for streaming receiver returns a `receiveHandler`, which has stop() method;
- receive() for streaming receiver **with sessions** returns `Void`.

**Consequences:**
- ReceiveBatch to clean the messages can't be called because the session is still active.

   - > Error: MessageSession 'standard5_session-e8beb792-f397-1b42-8c78-4869c75aa9c6' with sessionId 'my-session' is already receiving messages.

- The message is repeatedly received and before going to the Deadletter queue(until DeliveryCount reaches 10).

ramya-rao-a:
As discussed offline, to stop the streaming receiver, we need to call `close()` on the receiver client. That should fix the problem with the test cases for the abandon case

The one about deferred case is being tracked in https://github.com/Azure/azure-service-bus-node/issues/212

Since these tests are in a separate file altogether, it shouldnt be hard to create a new PR in the new repo.
Therefore, I'll close this PR.

PR #229, opened 2019-01-22
Resolved receive batch session error.

ShivangiReja:
* Removed `_notifyError(error)` (notify the user's error handler about the error.) method from `resetTimerOnNewMessageReceived` in `receiveBatch()`.
* Removed  `_resetTimerOnNewMessageReceived()` method and used    `resetTimerOnNewMessageReceived` as a constant for `receive()` and `receiveBatch()` method separately.


ShivangiReja:
Added similar PR here:  https://github.com/Azure/azure-sdk-for-js/pull/976

